PR #33, opened 2025-02-09
Feat/port councillor data population

mattorchard:
Ports and updates code to populate councillor and vote data
(hoping to update and merge saturday)


src/logic/toSlug.ts
```diff
@@ -0,0 +1,13 @@
...
+// eslint-disable-next-line no-restricted-imports
+import slugify from 'slugify';
+
+const slugifyOptions = { lower: true };
+
+export const toSlug = (input: string): string => {
+  if (!input) throw new Error(`Cannot slugify empty string`);
+  if (!input.trim()) throw new Error(`Cannot slugify pure whitespace`);
+  const output = slugify(input, slugifyOptions);
+  if (!output) throw new Error(`Slugifying "${input}" produced empty string`);
```
shrouxm, line 11: nit: prefer explicit checks
```suggestion
  if (input === '') throw new Error(`Cannot slugify empty string`);
  if (input.trim() === '') throw new Error(`Cannot slugify pure whitespace`);
  const output = slugify(input, slugifyOptions);
  if (output === '') throw new Error(`Slugifying "${input}" produced empty string`);
```

src/database/pipelines/textParseUtils.ts
```diff
@@ -0,0 +1,80 @@
+import { toSlug } from '@/logic/toSlug';
+
+/**
+ * Finds a term e.g. "2024-2026" within a larger string
+ * Throws if there is no term, or more than one term within the text
+ * @param text to search within e.g. "Lorem ipsum 2024-2026 dolor isut"
+ * @returns a term string e.g. "2024-2026"
+ */
+export function extractTermFromText(text: string): string {
```
shrouxm, line 9: comment: gross, thank u for sloshing thru this lol

src/app/api/councillors/refresh/route.ts
```diff
@@ -1,19 +1,8 @@
-import { db } from '@/database/kyselyDb';
```
shrouxm, line 1: comment: not immediately important, but i had the thought that if the cron jobs are run this way (trigger an API route with a secret), and it's a long running cron, it'll be using up resources the whole time. since we'll still need to run some compute somewhere to actually hit the API route, it could be better to also just run the cron task on that compute?

shrouxm (approved):
LGTM! couple small comments. i didn't really go over the CSV parsing stuff carefully since it seems like the fact that we're passing it thru kysely should at least take care of any security issues there, and any bugs/failures would just result in stale data rather than any more serious user-facing issues

src/app/api/councillors/refresh/route.ts
mattorchard: Yeah my thinking was purely that this is a good enough way to make it runnable for now without fracturing the TSconfig in a way that ts-node kinda ends up doing.
And eventually data team swoops in and solves this problem entirely.

.eslintrc.json
```diff
@@ -19,6 +19,10 @@
           {
             "name": "@/database/generatedDbTypes",
             "message": "Please use @/database/allDbTypes instead."
+          },
+          {
+            "name": "slugify",
+            "message": "Please use @/logic/toSlug instead."
```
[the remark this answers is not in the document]
mattorchard: Relatedly. I am wondering if it is worth it to have a dedicated eslint plugin for fixing typos of `councillor` since those are really easy to let sneak in on accident, and if it has an auto-fix it can be a solved problem.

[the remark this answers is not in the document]
shrouxm: sounds nice & like it wouldn't introduce any maintenance burden!

src/app/api/councillors/refresh/route.ts
shrouxm: :+1: okie, i'll see how things go once i start implementing email subs cuz that will be very cron-y

mattorchard:
Package lock is getting borked when I generate it locally. 
I might bug someone to try and regen on their device to see if that keeps the GH-Actions happy

PR #34, opened 2025-02-12
fix: don't initialise DB at top level

shrouxm:
not sure if this will work til it runs in CI, but I'm hoping this will fix the mystery 500 errors


shrouxm:
superseded by #35

PR #35, opened 2025-02-13
feat: follow-on DX suggestions from having added a database

shrouxm:
- lazily initializes the kysely DB instance (i think this fixes the weird 500 errors)
- adds down migrations for existing migrations
- adds a `.env.sample` file so it's easier to see which env vars need to exist for the app to run
- adds an eslint plugin for sql template strings to encourage consistency in query formatting
  - SQL is probably my least favorite language to auto-format, but i still prefer having the auto-formatting for the sake of it
- adds a `docker-compose.yml` file to start a postgres database for local development and CI checks
- adds `npm run db:{start,stop}` scripts to run that docker compose file
- adds a script which checks that the database migrations can all run and rollback successfully, and that the generated DB types are up to date
- changes kysely config to no longer log all queries (it's quite noisy, and individual developers can enable it locally if needed)

let me know if you disagree that any of these changes will improve the DX!


src/migrations/1737245494496_AddRawCouncillorTables.ts
```diff
@@ -45,3 +45,11 @@ export async function up(db: Kysely<unknown>): Promise<void> {
     );
   `.execute(db);
 }
+
+export async function down(db: Kysely<unknown>): Promise<void> {
```
mattorchard, line 49: I am curious why we want down migrations?
Over the years I have personally never found a time when I wanted them, as in cases where I did want to undo the result of a migration that was a change that I wanted to fix-forward as it was almost always some circumstance unforeseeable at the time the migration was written

mattorchard: This article's tone is maybe a little aggro, but expresses a lot of similar points
https://antman-does-software.com/why-you-will-never-write-another-down-migration

docker-compose.yml
```diff
@@ -0,0 +1,14 @@
+services:
+  database:
+    image: postgres:17.2
+
+    ports:
+      - 5432:5432
```
mattorchard, line 6: Is it worth aliasing the host port to avoid it failing to bind for anyone who has PSQL natively installed
```suggestion
      - 54320:5432
```

.github/workflows/check_migrations.yml
mattorchard, line 1: Really really nice

package.json
```diff
@@ -53,8 +55,10 @@
     "eslint-config-prettier": "^9.1.0",
     "eslint-plugin-prettier": "^5.2.1",
     "eslint-plugin-react": "^7.37.2",
+    "eslint-plugin-sql": "^3.2.1",
     "kysely-codegen": "^0.17.0",
     "kysely-ctl": "^0.10.1",
+    "pg": "^8.13.2",
```
mattorchard, line 61: Is this a peer dep of the lint plugin?

.env.sample
mattorchard, line 1: Possible alternative: we could have this info in the readme file, just to minimize chances someone modifies this file and then pushes creds

src/migrations/1737245494496_AddRawCouncillorTables.ts
shrouxm: i too have never run a down migration in prod, but i use them really often! thoughts:
- i mainly like having them for dev and staging databases
- if e.g. i'm working on two branches in parallel that both have migrations, it's nice to be able to quickly swap between them with the down migrations without corrupting my DB's migration table
- and similarly if there's a staging database for test deploys that's going back and forth between different users
- these needs could be addressed differently by making databases really easy to set up and tear down, but:
  - that takes work
  - IMO having a persistent staging database, which accumulates weird cruft over time like prod does, is really helpful for catching weird prod bugs in a way that constantly creating fresh ones can't compare to
- i take the point about down migrations not being tested, but:
  - it's not super hard to do at least some smoke testing of them (e.g. you can automatically check that the database schema is identical after running an up/down migration), it just needs to be set up the once
  - even if it's not how you end up addressing a prod bug, it is still really nice to have some code to start with when things are on fire
- one thing that has been tipping me more in favor of down migrations lately is that they take way less work than they used to, because it's the sort of thing LLMs really excel at
- it's easier to stop having down migrations than to start having down migrations

having written all that out, i can see that:
- it's unclear when/if we will have a staging DB
- i can always use them myself for local testing and just not commit them
- i can also see how it's more burden/overhead especially for occasional contributors

sooo yeah lmk what you think, if you're still meh i'm open to taking them out

docker-compose.yml
shrouxm: sure!

package.json
shrouxm: IIRC it's a peer dep of `kysely-codegen`!

.env.sample
shrouxm: fair! i will try to find a good place for it

mattorchard (approved):

src/migrations/1737245494496_AddRawCouncillorTables.ts
mattorchard: Ahh, I have not heard these points before, but they form a pretty bullet proof rationale. I'm on board!

For cases where a down migration is impossible (like when the up deletes a not-null column) is it correct to omit the down migration, or to have it and make it always throw?

shrouxm: @mattorchard if it's truly impossible i would definitely think throw! otherwise you're creating a situation where someone can silently end up in a totally untested schema

PR #37, opened 2025-02-20
Feat/add councillor api content

mattorchard:
Grabs all data for the current term from the council API and dumps the full content into a simple JSONB table.
Runs the requests for each page of the data in parallel, so it's lightning fast, but probably a bit spammy so probably shouldn't be run constantly


mattorchard:
Closing as councillor API now injested by Garo's pipeline

PR #38, opened 2025-02-20
feat: create agenda item table, and automatically keep it up to date

shrouxm:
- add new table `AgendaItem`
  - it has a unique constraint on `(reference, meetingId)` to support updates on conflict when refetching the same agenda items, so that updating agenda items can be idempotent
  - it is just an extremely literal translation of the data we get back from TMMIS API right now, there is no preprocessing step
- make functions for fetching agenda items from TMMIS API more flexible
- create some scripts to do initial population and ongoing updates
- schedule the update script for once an hour
- also runs the DB migrations on merge, curious people's thoughts on this or if we always want that to be manual ( :exclamation: )
  - i need to add a line so that it only runs on merge, not during the cron job

some things that are now ugh:
- the upsert is gnarly, but couldn't figure out how to simplify
  - it might end up being nicer to just pull the rows and do the deduplication logic in the typescript layer, especially once we start considering things like checking whether post-processed data needs updating, or flagging for email updates
- there is now an `AgendaItem` table, an `AgendaItems` view, and a `ProblemAgendaItems` view :thinking: 

Here is a compressed csv of all the agenda items for the current council term, so you can load it into your dev DB and start playing around right away: [currentTermAgendaItems.zip](https://github.com/user-attachments/files/18896629/currentTermAgendaItems.zip)

### DIFF
--- a/src/api/agendaItem.ts
+++ b/src/api/agendaItem.ts
@@ -61,23 +61,70 @@ export interface AgendaItem {
   neighbourhoodId?: number[];
 }
 
-export const fetchItems = async () => {
-  const now = new Date();
-  const nextMonth = new Date();
-  nextMonth.setMonth(nextMonth.getMonth() + 1);
+type AgendaItemFetchOptions = {
+  start: Date;
+  end: Date;
+};
+
+const fetchItemPage = async (
+  page: number,
+  { start, end }: AgendaItemFetchOptions,
+) => {
   const body = JSON.stringify({
     includeTitle: true,
     includeSummary: true,
     includeRecommendations: true,
     includeDecisions: true,
-    meetingFromDate: now.toISOString(),
-    meetingToDate: nextMonth.toISOString(),
+    meetingFromDate: start.toISOString(),
+    meetingToDate: end.toISOString(),
   });
 
   const response = await cityCouncilXSRFPost({
-    url: 'https://secure.toronto.ca/council/api/multiple/agenda-items.json?pageNumber=0&pageSize=200&sortOrder=meetingDate%20asc,referenceSort',
+    url: `https://secure.toronto.ca/council/api/multiple/agenda-items.json?pageNumber=${page}&pageSize=200&sortOrder=meetingDate%20asc,referenceSort`,
     body,
   });
 
-  return ((await response.json()) as ApiResponse).Records;
+  return (await response.json()) as ApiResponse;
+};
+
+const deduplicateAgendaItems = (items: AgendaItem[]) => {
+  const seen = new Map<string, AgendaItem>();
+
+  items.forEach((item) => {
+    const key = `${item.reference}-${item.meetingId}`;
+    const prev = seen.get(key);
+    if (prev === undefined || item.agendaItemId >= prev.agendaItemId) {
+      seen.set(key, item);
+      if (prev !== undefined) {
+        console.log(
+          `ignoring ${prev.agendaItemId} in favor of ${item.agendaItemId}`,
+        );
+      }
+    } else {
+      console.log(
+        `ignoring ${item.agendaItemId} in favor of ${prev.agendaItemId}`,
+      );
+    }
+  });
+
+  return [...seen.values()];
+};
+
+export const fetchAgendaItems = async (opts: AgendaItemFetchOptions) => {
+  let pageNumber = 0;
+  let totalPages = 1;
+  const records: AgendaItem[] = [];
+
+  while (pageNumber < totalPages) {
+    const page = await fetchItemPage(pageNumber, opts);
+    records.push(...page.Records);
+    totalPages = page.TotalPages;
+    pageNumber++;
+    if (pageNumber < totalPages) {
+      // wait a second before fetching again to avoid overloading the TMMIS API
+      await new Promise((resolve) => setTimeout(resolve, 1000));
+    }
+  }
+
+  return deduplicateAgendaItems(records);
 };
--- a/src/database/columns.ts
+++ b/src/database/columns.ts
@@ -1,17 +1,6 @@
-import { AgendaItem } from '@/database/allDbTypes';
+import { RawAgendaItemConsiderations } from '@/database/allDbTypes';
 
-type Assert<_T extends true> = never;
-type Equivalent<A extends string, B extends string> = A extends B
-  ? B extends A
-    ? true
-    : `${A} does not match ${B}`
-  : `${B} does not match ${A}`;
-type ColumnsMatch<A extends object, B extends readonly string[]> = Equivalent<
-  keyof A & string,
-  B[number]
->;
-
-export const agendaItemColumns = [
+export const agendaItemConflictColumns = [
   'address',
   'agendaCd',
   'agendaItemAddress',
@@ -39,8 +28,4 @@ export const agendaItemColumns = [
   'termId',
   'termYear',
   'wardId',
-] as const;
-
-type _AssertAgendaItemColumnsComplete = Assert<
-  ColumnsMatch<AgendaItem, typeof agendaItemColumns>
->;
+] as const satisfies (keyof RawAgendaItemConsiderations)[];
--- a/src/database/columns.ts
+++ b/src/database/columns.ts
@@ -1,0 +1,46 @@
+import { AgendaItem } from '@/database/allDbTypes';
+
+type Assert<_T extends true> = never;
+type Equivalent<A extends string, B extends string> = A extends B
+  ? B extends A
+    ? true
+    : `${A} does not match ${B}`
+  : `${B} does not match ${A}`;
+type ColumnsMatch<A extends object, B extends readonly string[]> = Equivalent<
+  keyof A & string,
+  B[number]
+>;
+
+export const agendaItemColumns = [
+  'address',
+  'agendaCd',
+  'agendaItemAddress',
+  'agendaItemId',
+  'agendaItemRecommendation',
+  'agendaItemSummary',
+  'agendaItemTitle',
+  'backgroundAttachmentId',
+  'councilAgendaItemId',
+  'decisionAdvice',
+  'decisionBodyId',
+  'decisionBodyName',
+  'decisionRecommendations',
+  'geoLocation',
+  'id',
+  'itemProcessId',
+  'itemStatus',
+  'meetingDate',
+  'meetingId',
+  'meetingNumber',
+  'neighbourhoodId',
+  'planningApplicationNumber',
+  'reference',
+  'subjectTerms',
+  'termId',
+  'termYear',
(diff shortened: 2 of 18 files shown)


src/scripts/populateDatabase.ts
```diff
@@ -0,0 +1,10 @@
+import { populateAgendaItems } from '@/database/pipelines/populateAgendaItems';
+
+const beginningOfTerm = 883630800000;
+const endOfTerm = 1794632400000;
+
+await populateAgendaItems(new Date(beginningOfTerm), new Date(endOfTerm));
+
+// for some reason the script doesn't end properly when run for a wide date range
+// so manually exit here
+process.exit(0);
```
mattorchard, line 10: Does the DB connection need to be explicitly killed to get it to yield any open handles?

shrouxm: oh HMM could be. i'm confused why it would only trigger for long-running scripts in that case (if i just do a couple months at a time it exits fine), but still a possible candidate. that's annoying to have to remember to do :/ also how does that work in the context of a next request?

src/database/columns.ts
```diff
@@ -0,0 +1,46 @@
...
+  'reference',
+  'subjectTerms',
+  'termId',
+  'termYear',
+  'wardId',
+] as const;
+
+type _AssertAgendaItemColumnsComplete = Assert<
+  ColumnsMatch<AgendaItem, typeof agendaItemColumns>
+>;
```
mattorchard, line 46: Does the assert give us additional checking what `satisfies` can do?
```suggestion
] as const satisfies ReadonlyArray<keyof AgendaItem>;
```

shrouxm: satisfies will guarantee that all strings in the array are keys of the object, but it _won't_ guarantee that all keys of the object are in the array. so e.g. removing `'wardId'` would error with this gross typescript nonsense i wrote, but not with the nice clean typescript suggestion you made :shrug:. it's a bit overwrought and i'm happy to remove it, i just like types and programming languages so typescript nonsense makes me happy (until it makes me sad)

src/scripts/populateDatabase.ts
mattorchard: In the context of a next req it's gonna depend a bit on the runtime and compute paradigm.
When run in a docker image it's straightforward cause the pool object just exists and manages it's own connection lifecycles. In most serverless runtimes the runtime can get killed anytime after the response is provided so open handles aren't even guaranteed to finish. All that said...I've never used workerd before so have 0 clue lol

mattorchard: The multiple months thing could be tied to something internal to the psql library trying to be clever about how long a connection should last based on the observed query frequency

src/database/columns.ts
mattorchard: I also love writing elaborate TS types, I think it scratches a similar itch to super compact regexes.
Both are also much more fun to write than to read/debug 😆 

If we wanna have a utility types folder we could add a few of these things and add some explanations to make it more approachable.

src/migrations/1740007960791_AddAgendaItems.ts
```diff
@@ -0,0 +1,92 @@
...
+      "agendaItemRecommendation" TEXT,
+      "decisionRecommendations" TEXT,
+      "decisionAdvice" TEXT,
+      "subjectTerms" TEXT NOT NULL,
+      "wardId" json,
+      "backgroundAttachmentId" json,
+      "agendaItemAddress" json,
+      "address" json,
+      "geoLocation" json,
+      "planningApplicationNumber" TEXT,
```
mattorchard, line 31: Any reason to prefer using `json` instead of `int[]` for the arrays of IDs? (it would give us easier access to things like any/all)

src/database/pipelines/populateAgendaItems.ts
```diff
@@ -0,0 +1,27 @@
...
+      start,
+      end: nextMonth > end ? end : nextMonth,
+    });
+    if (agendaItems.length > 0) {
+      const result = await insertAgendaItems(agendaItems);
+      insertedCount += result[0].numInsertedOrUpdatedRows ?? BigInt(0);
+    }
+    console.log('inserted:', insertedCount);
+
+    start = nextMonth;
```
mattorchard, line 20: Can we avoid reassigning param by having a `let current = start` up top and `current = nextMonth` here

src/api/agendaItem.ts
```diff
@@ -61,23 +61,70 @@ export interface AgendaItem {
...
+
+  while (pageNumber < totalPages) {
+    const page = await fetchItemPage(pageNumber, opts);
+    records.push(...page.Records);
+    totalPages = page.TotalPages;
+    pageNumber++;
+    if (pageNumber < totalPages) {
+      // wait a second before fetching again to avoid overloading the TMMIS API
+      await new Promise((resolve) => setTimeout(resolve, 1000));
+    }
```
mattorchard, line 126: We can avoid duplicating the comment and impl by pulling into a lil util 
```ts
      await artificialDelayToPreventSpamming()
```

src/database/pipelines/populateAgendaItems.ts
```diff
@@ -0,0 +1,27 @@
+import { fetchAgendaItems } from '@/api/agendaItem';
+import { insertAgendaItems } from '@/database/queries/agendaItems';
+
+export const populateAgendaItems = async (start: Date, end: Date) => {
+  let insertedCount = BigInt(0);
+  while (start < end) {
```
mattorchard, line 6: Why do we need to go through each month one at a time if the internal helper is already doing one page at a time?

src/api/agendaItem.ts
```diff
@@ -61,23 +61,70 @@ export interface AgendaItem {
...
+  return [...seen.values()];
+};
+
+export const fetchAgendaItems = async (opts: AgendaItemFetchOptions) => {
+  let pageNumber = 0;
+  let totalPages = 1;
+  const records: AgendaItem[] = [];
+
+  while (pageNumber < totalPages) {
+    const page = await fetchItemPage(pageNumber, opts);
```
mattorchard, line 119: In future we could use a client wrapper to hold onto the CSRF to cut our requests in half
(The experimental version I had in the old repo took that approach so there's a [ref impl](https://github.com/civic-dashboard/civic-dashboard-web/pull/37/files#diff-c7ad86dda7418aa67c831f467da0b4861e4b08437cb1a9abecdc717eaf233968))

src/database/queries/agendaItems.ts
```diff
@@ -0,0 +1,38 @@
...
+                .case()
+                .when(
+                  'AgendaItem.agendaItemId',
+                  '<=',
+                  eb.ref('excluded.agendaItemId'),
+                )
+                .then(eb.ref(`excluded.${k}`))
+                .else(eb.ref(`AgendaItem.${k}`))
+                .end(),
+            ]),
```
mattorchard, line 32: Since query builders and expression builders are already getting short-handed I think using a full and consistent name here would help readaibility
```suggestion
            .filter((column) => column !== 'id')
            .map((column) => [
              column,
              eb
                .case()
                .when(
                  'AgendaItem.agendaItemId',
                  '<=',
                  eb.ref('excluded.agendaItemId'),
                )
                .then(eb.ref(`excluded.${column}`))
                .else(eb.ref(`AgendaItem.${column}`))
                .end(),
            ]),
```

src/scripts/populateDatabase.ts
```diff
@@ -0,0 +1,10 @@
+import { populateAgendaItems } from '@/database/pipelines/populateAgendaItems';
+
+const beginningOfTerm = 883630800000;
```
mattorchard, line 3: I assume this works, but is definitely the kinda thing that makes me nervous about timezones potentially messing with.
Maybe in a future pass we could use a DateStruct instead of the built in Date obj just to avoid potential headaches

src/migrations/1740007960791_AddAgendaItems.ts
```diff
@@ -0,0 +1,92 @@
+import { sql, type Kysely } from 'kysely';
+
+export async function up(db: Kysely<unknown>): Promise<void> {
+  await sql`
+    CREATE TABLE "AgendaItem" (
```
mattorchard, line 5: > there is now an AgendaItem table, an AgendaItems view, and a ProblemAgendaItems view 🤔
Since the rows in this table are many-to-one with agenda items could we consider naming them after something they are 1-1 with.
Maybe `RawMeetings`?

src/scripts/populateDatabase.ts
shrouxm: hrm yeah, this is my first time working with any kind of JS backend server stuff so i'm not even sure what the usual pattern would be. i'm used to passing around transaction objects or using some mechanism that can wrap a scope in a transaction rather than a global database ref. but it sounds like this is not blocking for the PR?

src/migrations/1740007960791_AddAgendaItems.ts
shrouxm: norp!

src/database/pipelines/populateAgendaItems.ts
shrouxm: i'm not sure any of the usual reasons i try to avoid doing massive queries by pagination apply here, but some general reasons:
- the page number is a happenstance property of the data which could change at any moment, whereas a static property like the date ought to be more consistent. e.g. if i run the script once today, and it fails halfway thru, and i want to run it again without re-running it over the old data, i have no guarantee the page numbers will still line up properly, but the dates should still be good
- depending on how the database is set up and the exact query situation, asking for the full data set and going thru it one page at a time could trigger a full data scan on each query in a way that asking for a specific slice of data might not

src/api/agendaItem.ts
shrouxm: agree! i was just too lazy to pass a client around and don't like having stateful global instances of things

src/migrations/1740007960791_AddAgendaItems.ts
shrouxm: i believe the domain term for what's in the table is a "consideration", but naming the table `Consideration` feels confusing. it could be `AgendaItemConsideration`, which is a bit of a mouthful but that's ok. i definitely don't want to do `Meeting` because a meeting encompasses a full agenda of items.

does the `AgendaItems` view show multiple entries if an agenda item is considered in multiple meetings, or is there just one entry per agenda item reference?

mattorchard: Right now the agenda items view is 1-1 with the agenda item number, regardless of how many meetings or motions come out of it.

mattorchard (approved):

PR #41, opened 2025-03-01
Eslint rule to prevent typoing councillor

mattorchard:
A little bit jank to be installing itself locally.
But seemingly wasn't able to get eslint `8.57` happy any other way with a locally defined plugin.


mattorchard:
Fun but a nightmare to maintain, and ultimately not that important

PR #51, opened 2025-03-11
create how-council-works page, tooltip component

sarmilan:


src/app/how-council-works/page.tsx
```diff
@@ -0,0 +1,257 @@
...
+            <span className="font-bold">Submit a Comment</span>
+            <br />
+            At the committee stage, you can also make a comment on each agenda
+            item that is being considered. Your comments can influence the
+            members of the committee as they make their decisions.
+            <br />
+            <br />
+            Want more ways to get involved, or how to raise an unaddressed
+            issue? Visit the{' '}
+            <a href="/coming-soon" className="text-[#0073e6] underline">
```
mattorchard, line 248: If this color is gonna keep showin up: should we componentize / add to our tailwind config?

src/components/ui/tooltip.tsx
```diff
@@ -0,0 +1,64 @@
...
+            sideOffset={4}
+            className="z-50 w-[400px] border border-[#E0E0E0] overflow-hidden rounded-md bg-white p-[23px] text-black animate-in fade-in-0 zoom-in-95 data-[state=closed]:animate-out data-[state=closed]:fade-out-0 data-[state=closed]:zoom-out-95 data-[side=bottom]:slide-in-from-top-2 data-[side=left]:slide-in-from-right-2 data-[side=right]:slide-in-from-left-2 data-[side=top]:slide-in-from-bottom-2"
+          >
+            <div className="relative">
+              <button
+                onClick={() => setIsOpen(false)}
+                className="absolute -right-[11px] -top-[11px] rounded-full w-6 h-6 flex items-center justify-center text-gray-500 hover:text-gray-700"
+                aria-label="Close tooltip"
+              >
+                ✕
```
mattorchard, line 52: Could we use the lucide x icon?

src/components/ui/tooltip.tsx
```diff
@@ -0,0 +1,64 @@
...
+      'ontouchstart' in window || navigator.maxTouchPoints > 0;
+  }, []);
+
+  return (
+    <TooltipPrimitive.Provider>
+      <TooltipPrimitive.Root open={isOpen} onOpenChange={setIsOpen}>
+        <TooltipPrimitive.Trigger
+          asChild
+          onClick={(e) => {
+            if (isTouchDevice.current) {
```
mattorchard, line 31: Does radix not properly support mobile?

I believe this condition will be true for devices that are not touch-only like windows surfaces, or really anything witha touchscreen even if it also has a regular mouse. Is that expectde?

src/app/how-council-works/page.tsx
```diff
@@ -0,0 +1,257 @@
+'use client';
+
+import { TooltipProvider } from '@radix-ui/react-tooltip';
```
mattorchard, line 3: If we re-export this in the tooltip component we can import them beside one another

src/app/how-council-works/page.tsx
sarmilan: this is me being a lazy boi - I'll update the a tag styling in global.css since the colour isn't important

src/components/ui/tooltip.tsx
sarmilan: ✅

src/app/globals.css
```diff
@@ -28,6 +28,10 @@
   :root {
     --radius: 0.5rem;
   }
+
+  main a {
```
mattorchard, line 32: Should we still restrict this to a className just to keep it opt in?
```suggestion
  .classic-link {
```

src/app/how-council-works/page.tsx
```diff
@@ -0,0 +1,243 @@
...
+            decisions follows three key steps.
+          </div>
+
+          <div className="w-full h-[540px] mt-14 border border-neutral-400 flex items-center justify-center">
+            <span className="text-neutral-500">Image Placeholder</span>
+          </div>
+        </div>
+        {/* Three-Step City Council Process */}
+        <div className="mt-[78px]">
+          <h2 className="text-[40px] font-bold text-center text-[24px] md:text-[32px] lg:text-[40px]">
```
mattorchard, line 63:
```suggestion
          <h2 className="font-bold text-center text-[24px] md:text-[32px] lg:text-[40px]">
```

src/app/how-council-works/page.tsx
```diff
@@ -0,0 +1,243 @@
...
+            The Three-Step City Council Process
+          </h2>
+          <div className="text-[14px] md:text-[16px] lg:text-[18px] leading-7 mt-8 mt-[80px] max-w-2xl mx-auto">
```
mattorchard, line 66:
```suggestion
          <div className="text-[14px] md:text-[16px] lg:text-[18px] leading-7 mt-8 max-w-2xl mx-auto">
```

src/components/ui/tooltip.tsx
```diff
@@ -0,0 +1,67 @@
...
+  tooltipContent,
+  tooltipTitle,
+}: TooltipProps) {
+  const [isOpen, setIsOpen] = React.useState(false);
+  const isTouchDevice = React.useRef(false);
+
+  React.useEffect(() => {
+    isTouchDevice.current =
+      'ontouchstart' in window || navigator.maxTouchPoints > 0;
+  }, []);
```
mattorchard, line 24: Not a deal breaker but the condition could just be checked by a helper during the onclick instead of ref+effect

src/app/how-council-works/page.tsx
```diff
@@ -0,0 +1,243 @@
...
+            <br />
+            <br />
+            For a more in-depth explanation, visit the{' '}
+            <a href="/coming-soon">City Council Wiki</a>.
+          </div>
+        </div>
+
+        {/* Get Involved */}
+        <div className="mt-[78px] mb-20">
+          <h2 className="text-[24px] md:text-[32px] lg:text-[40px] font-bold text-center">
```
mattorchard, line 194: Not a huge deal but having a couple one-liner components defined in the same file to handle adding the same heading and section classes might make it easier to distinguish between content and code when reading

(also I think a lot of these measurements are regular spacing numbers in tailwind)

sarmilan: good idea! for the tailwind spacing numbers, I'll keep it as pixels since the specs are in pixels and its easier to ensure specs

mattorchard (approved):

Issue #54, opened 2025-03-12
Add instructions on how to load data locally in a DEV environment


jeromegv:

jeromegv:
Assigned to me.

jeromegv:
@shrouxm can you add me to the repo so that i can create a branch/PR? thank you!

shrouxm:
@jeromegv are you able to now?

jeromegv:
yep all good! @shrouxm

shrouxm:
compelted in #65

PR #65, opened 2025-03-17
Docs update README to add info on how to load data

jeromegv:
#54


jeromegv:
@shrouxm its a small one, can we merge? I have #66 for instructions on how to load the full thing instead

PR #67, opened 2025-03-17
Make dom-safe temporary summaries available

mattorchard:


package.json
```diff
@@ -13,6 +13,7 @@
     "preview-local:worker": "npm run build:worker && npm run dev:worker",
     "preview-remote:worker": "npm run build:worker && wrangler versions upload",
     "deploy:worker": "npm run build:worker && wrangler deploy",
+    "tsx": "tsx",
```
mattorchard, line 16: Seemingly needed to let windows run the tsx pipelines

shrouxm: i added a `tsx` command at the bottom for an unrelated reason, so this should hopefully be unnecessary!

shrouxm: / might break the email update script!

mattorchard: ty for catching that, the file didn't give any merge conflicts. I guess on account of the space between

shrouxm: sg

PR #74, opened 2025-03-21
Feat: infinite scroll on actions page

### LINKED ISSUE #47: Implement infinite scroll/paging for actions page
Currently we just fetch the first 50 items from the API and that's it. We should instead fetch a smaller number (10-20), and load more when the user scrolls to the bottom of the list.

luceliu:
### Infinite scroll
* Adds a `useEffect` to `SearchContext` to trigger another fetch request when `currentPage` increments
* Modifies `onSearch` to take a `page` param, and make fetch requests with a default page limit of `10`
* Adds `useInfiniteScroll` in new `src/hooks` folder
* Installs a new dependency `react-intersection-observer` and wraps it in `src/hooks/useIntersectionObserver.ts`

### Other refactors
* Splits `const [searchResults, setSearchResults] = useState<AgendaItemSearchResponse | 'loading'>('loading');` into 2 separate states:
     * `const [searchResults, setSearchResults] = useState<AgendaItemSearchResponse | null>(null);`
     * `const [isLoading, setIsLoading] = useState<boolean>(true);`
* Extracts search context logic into a new `src/contexts/SearchContext.tsx`
* Extracts some search related constants into a new `src/constants/search.ts`

Closes #47 

I took the liberty of creating some new files/folders -- let me know if this is an issue! I had hoped to be able to pull more of the infinite scrolling logic into `useInfiniteScroll` to not have it peppered throughout `SearchProvider` but that turned out to be more gnarly than expected, at least the way I was trying to do it, so maybe a potential refactor for a future PR unless y'all have thoughts? Very open to feedback in general :-)


shrouxm (approved):
omg this is so clean :heart_eyes: gonna merge right now, ty!

PR #76, opened 2025-03-22
Councillor list page updates

mattorchard:
- Adds filter
- Adds second column
- Updates link to a better councillor link
- Uses Next image instead of regular image and swaps to round avatars


src/components/SearchInput.tsx
mattorchard, line 1: At some point we should make our different filter stuff behave similarly, but that seems like a future problem

src/components/Avatar.tsx
mattorchard, line 1: Didn't add this to councillor bio yet since I didnt wanna create potential conflicts with Milan's work

PR #82, opened 2025-03-28
create and call notifcationBanner with feedback message

sarmilan:
https://www.loom.com/share/42a92d455e2e4287bf8fe9c6cf43feef?sid=8cf8e585-af2d-4244-bfa3-c40d23026a57

### DIFF
--- a/src/components/navigation/NotificationBanner.tsx
+++ b/src/components/navigation/NotificationBanner.tsx
@@ -1,0 +1,51 @@
+'use client';
+import { useState } from 'react';
+import { useRouter } from 'next/navigation';
+
+export default function NotificationBanner({
+  message,
+  link,
+}: {
+  message: string;
+  link: string;
+}) {
+  const [isCollapsed, setIsCollapsed] = useState(false);
+  const router = useRouter();
+
+  const handleBannerClick = () => {
+    router.push(link);
+  };
+
+  return (
+    <div className="fixed w-full top-16 z-50">
+      {!isCollapsed ? (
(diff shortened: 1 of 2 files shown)


src/components/navigation/NotificationBanner.tsx
```diff
@@ -0,0 +1,51 @@
...
+
+  return (
+    <div className="fixed w-full top-16 z-50">
+      {!isCollapsed ? (
+        <div className="bg-orange-400 text-white w-full">
+          <div
+            className="max-w-7xl mx-auto px-4 sm:px-6 lg:px-8 py-2 flex items-center justify-between cursor-pointer"
+            onClick={handleBannerClick}
+          >
+            <p className="text-md font-semibold">{message}</p>
```
mattorchard, line 27: If we use a Link we can make this accessible and keyboard-usable

src/components/navigation/NotificationBanner.tsx
```diff
@@ -0,0 +1,51 @@
...
+import { useRouter } from 'next/navigation';
+
+export default function NotificationBanner({
+  message,
+  link,
+}: {
+  message: string;
+  link: string;
+}) {
+  const [isCollapsed, setIsCollapsed] = useState(false);
```
mattorchard, line 12: Should we start collapsed or hide this entire component when we are already on the feedback page so that it's clearer the user has landed where they need to be?

sarmilan: ah, good idea

mattorchard (approved):

PR #83, opened 2025-03-28
update tooltip definitions, placements

sarmilan:


src/app/how-council-works/page.tsx
```diff
@@ -256,6 +302,26 @@ export default function HowCouncilWorks() {
             </span>
           </SectionText>
         </Section>
+        <Section>
+          <SectionHeading>Stay Informed, Take Action</SectionHeading>
+          <SectionText>
+            Now that you know how City Council works, why not take the next
+            step? Stay informed and make your voice heard! Check out our{' '}
+            <a href={menuItems[3].href} className="classic-link">
```
mattorchard, line 310: I think aray-indexes into the nav end up being kinda fragile, since someone might reorder the array and then the link goes to something unrelated
(also we could use the `<Link` component here)
```suggestion
            <Link href="/actions" className="classic-link">
```

src/app/how-council-works/page.tsx
```diff
@@ -256,6 +302,26 @@ export default function HowCouncilWorks() {
...
+              Actions Page
+            </a>{' '}
+            to see upcoming Council items and take action in just one click.
+            <br />
+            <br />
+            Or visit our{' '}
+            <a href={menuItems[4].href} className="classic-link">
+              Councillors Page
+            </a>{' '}
```
mattorchard, line 319:
```suggestion
            <Link href="/councillors className="classic-link">
              Councillors Page
            </Link>{' '}
```

src/app/how-council-works/page.tsx
```diff
@@ -256,6 +302,26 @@ export default function HowCouncilWorks() {
...
           </SectionText>
         </Section>
+        <Section>
+          <SectionHeading>Stay Informed, Take Action</SectionHeading>
+          <SectionText>
+            Now that you know how City Council works, why not take the next
+            step? Stay informed and make your voice heard! Check out our{' '}
+            <a href={menuItems[3].href} className="classic-link">
+              Actions Page
+            </a>{' '}
```
mattorchard, line 312: I think aray-indexes into the nav end up being kinda fragile, since someone might reorder the array and then the link goes to something unrelated
(also we could use the `<Link` component here)
```suggestion
            <Link href="/actions" className="classic-link">
              Actions Page
            </Link>{' '}
```

src/app/how-council-works/page.tsx
```diff
@@ -41,31 +49,60 @@ export default function HowCouncilWorks() {
...
+      alt: 'Illustration of a staff meeting on affordable housing: a group of city staff sits around a table in an office discussing housing policy. Thought bubbles above them depict housing types, consultations, and paperwork.',
+    },
+    committee: {
+      src: '/committee.png',
+      alt: 'Illustration of a committee meeting on affordable housing: a speaker at a podium addresses a panel of business casual committee members, while diverse citizens in the foreground share their opinions via speech bubbles.',
+    },
+    council: {
+      src: '/council.png',
+      alt: 'Illustration of a City Council meeting on affordable housing: the councillors have voted on an item, with the result displayed on the screen showing it has been approved. Touchscreens in front of the councillors indicate votes with green for Approve and red for Reject.',
     },
```
mattorchard, line 100: Might actually be better to blank out the alt's here since I think these images fit the criteria for decorative/ambiance images
https://www.w3.org/WAI/tutorials/images/decorative/

src/app/how-council-works/page.tsx
```diff
@@ -256,6 +326,42 @@ export default function HowCouncilWorks() {
...
+            to see how your Councillor has voted and whether they align with
+            your priorities. Your engagement can help shape the decisions that
+            impact Toronto—get involved today!
+          </SectionText>
+        </Section>
+        <Section>
+          <SectionText>
+            <span className="italic">
+              Illustrations by Luisa Castillo Henao. View more of her work on{' '}
+              <Link
```
mattorchard, line 353: For this we should use `ExternalLink` that way it goes to a new tab and adds the rel option to prevent reverse tabnabbing

PR #84, opened 2025-03-30
Import ai summaries for councillor votes page

mattorchard:
Very much a temp solution to try and pull in AI summaries from an external CSV in advance of the initial launch

1. Adds migration to add an ultra basic table
2. Adds a pipeline to populate it with only the simplest cases

Draft PR shows how they could start to be used
https://github.com/civic-dashboard/civic-dashboard-web/pull/86


src/database/pipelines/importAiSummaries.ts
```diff
@@ -0,0 +1,92 @@
...
+  }
+
+  private async getCandidateIdLookup() {
+    const rows = await this.trx
+      .selectFrom('AgendaItems')
+      .innerJoin('RawAgendaItemConsiderations', (eb) =>
+        eb.onRef('agendaItemNumber', '=', 'reference'),
+      )
+      .groupBy('agendaItemNumber')
+      .having(sql`count(*)`, '=', 1)
```
mattorchard, line 53: Basically: only summarize cases where the considerations are 1-1 with agenda items

src/database/pipelines/importAiSummaries.ts
```diff
@@ -0,0 +1,92 @@
...
+  private async run() {
+    if (!this.inputFileName)
+      throw new Error(`Invalid file name ${this.inputFileName}`);
+
+    const candidateIdLookup = await this.getCandidateIdLookup();
+    const inputStream = await this.createInputStream();
+
+    const summaries = new Array<SummaryRow>();
+    for await (const inputRow of inputStream) {
+      const agendaItemNumber = candidateIdLookup.get(+inputRow.agendaItemId);
```
mattorchard, line 34: Ideally future CSVs will just include the agenda item number, instead of the internal ID. But this works for the existing file

src/database/pipelines/importAiSummaries.ts
mattorchard: Removed this restriction since the considerations may update on their own and break this assumption. So better for the consumer to handle it

.gitignore
```diff
@@ -48,3 +48,6 @@ next-env.d.ts
 
 # database
 db-data
+
+
+data/
```
shrouxm, line 53: nit: missing newline at end of file

src/database/pipelines/importAiSummaries.ts
```diff
@@ -0,0 +1,85 @@
...
```
shrouxm, line 34: question: why the `+`?

shrouxm (approved):
LGTM. i'm only now noticing we call them agenda item "number"s in some places and "reference"s in other places, which i'd love to get more consistent about, but don't think that needs to happen in this PR

src/database/pipelines/importAiSummaries.ts
mattorchard: yeah this line is cursed.
agendaItemNumber is the human readable ID for the agenda item (which appears in the TMMIS UI and ours and the URLs for the TMMIS API). But the `agendaItemID` is the internal numeric ID.
Since this is coming from a CSV tho, everything is a string.

So this line finds the `agendItemNumber: string` using the `agendaItemId: number` but needs to force that ID into a number from a string

shrouxm: ohhh, so is `+` basically acting as a call to `parseInt`/`Number` here? couuuld we use `parseInt` to be more explicit?

mattorchard: New CSV includes the reference column. So the cursed line in question is gone

PR #85, opened 2025-03-30
Swap open data pipeline to TSX instead of API route

mattorchard:


src/database/pipelines/repopulateRawContactsAndVotes.ts
```diff
@@ -147,3 +143,6 @@ async function* asBatches<T>(
   }
   if (batch.length > 0) yield batch;
 }
+
+await RepopulateRawContactsAndVotes.run(createDB());
+process.exit(0);
```
shrouxm, line 148: i've been putting the actual invocations of different scripts in `src/scripts/...`, so they can all be run as `npm run tsx src/scripts/...`. i'm not sure why i've been doing this, but it'd be nice to be consistent!

shrouxm (approved):
LGTM. one minor comment. i guess your plan is still to run it manually ad-hoc? you could integrate it into the `updateDatabase` script & then it'll get run every hour automatically

src/database/pipelines/repopulateRawContactsAndVotes.ts
mattorchard: Oh dang, missed that, will swap

mattorchard:
> LGTM. one minor comment. i guess your plan is still to run it manually ad-hoc? you could integrate it into the `updateDatabase` script & then it'll get run every hour automatically

For today I think makes sense to still say separate. The perf aint great when run against sevalla rn, and this dataset doesn't even seem to update weekly

PR #96, opened 2025-04-03
fix: shorter comments mailto

shrouxm:


shrouxm:
NVM, turns out duck duck go browser is just broken

PR #101, opened 2025-04-04
Add councillor init to readme

mattorchard:


package.json
```diff
@@ -21,7 +21,8 @@
     "db:stop": "docker compose stop database",
     "docker:start": "docker compose up --detach",
     "docker:stop": "docker compose stop",
-    "tsx": "tsx --tsconfig tsconfig.tsx.json"
+    "tsx": "tsx --tsconfig=tsconfig.tsx.json",
+    "tsxe": "tsx --tsconfig=tsconfig.tsx.json --env-file=.env"
```
mattorchard, line 25: I think having this is really nice, but I couldn't think of a name that made sense that wasn't painfully verbose.
(didn't wanna edit the original since node exits when it can't find it and running off actual env vars and not a file is probably what our GH actions rely on)

PR #138, opened 2025-05-13
update text colour classes to leverage tailwind darkmode

sarmilan:
Screenshot 2025-05-13 at 7 25 24 PM

### DIFF
--- a/src/app/about/page.tsx
+++ b/src/app/about/page.tsx
@@ -6,16 +6,14 @@ export const metadata: Metadata = {
 
 export default function About() {
   return (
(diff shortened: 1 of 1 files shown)


src/app/about/page.tsx
```diff
@@ -199,7 +191,7 @@ export default function About() {
                 <a
```
mattorchard, line 191: Could use `ExternalLink` component to dedupe target+rel

mattorchard: `src/components/ExternalLink.tsx`

mattorchard (approved):

Issue #127, opened 2025-04-30
Add tagging data to database for category filtering


melissaluc:
### Description

- Set up tables for the categories and tags in the database to enable filtering functionality on the councillor page.
- Populate the tables with the data from Khasir's investigation which can be found in a report[^1] and documented in notebooks[^2][^3][^4] with outputs generated from the Subject Term Grouping Notebook[^2].
     - **Key Analysis Outputs:**
         - all_terms.csv
         - histogram_table.csv
         - top_terms.csv
     - **Key Analysis Inputs:**
          - all_items/similarities.csv
          - all_items/comma_separated_subject_terms.txt
- More context can be drawn from the slack thread[^5] and the project brief on TMMIS tagging system[^6].

>[!NOTE] 
>Current Assumption(s):
>- The Subject Term Analysis expands on the TMMIS City council staff current tag system on agenda items[^7], breaking down existing categories into a hierarchy of sub category tags

### To-do (Acceptance Criteria)

- [x] Add TagsCategory table
- [x] Add AgendaItemCategories table
- [x] Add AgendaItemSubjectTerm table

[^1]: [Subject Term Analysis Report](https://docs.google.com/document/d/1vFlxKObVFz55KUnhoRGYPrFbOHN8-4XDa1s8IMYjJAc/edit?tab=t.0#heading=h.9ph2w63dwn1u)
[^2]: [Subject Term Grouping](https://github.com/Khasir/civic-dashboard/blob/25-02-11-subject-terms/ml/categories/notebooks/25-03-12-subject-term-grouping.ipynb)
[^3]: [Subject Term Keyword Exploration](https://github.com/Khasir/civic-dashboard/blob/25-02-11-subject-terms/ml/categories/notebooks/25-02-25-keyword-exploration.ipynb)
[^4]: [Subject Term Data Exploration](https://github.com/Khasir/civic-dashboard/blob/25-02-11-subject-terms/ml/categories/notebooks/25-02-11-data-exploration.ipynb)
[^5]: [Slack Message on City Council Agenda Subject Terms (2022–Present)](https://civictechto.slack.com/archives/C06KU3DHEKV/p1740092454564759)
[^6]: [TMMIS Google Docs](https://docs.google.com/document/d/1elazNTCHvhAoHuva8HdSStC6LKVTHdTwKUR7NIs9DyU/edit?tab=t.0#heading=h.a0ikyem60ndl)
[...]

PR #149, opened 2025-05-16
Migration tags data

melissaluc:
## Description

- Added migration and seed file for TagCategories and AgendaItemsCategories View

## Related issue

#127 

## Review Checklist:
- [ ] All to-do items (acceptance criteria) from ticket have been treated
- [x] Has the change been manually tested in the preview deployment?


src/migrations/1746652473628_AddAgendaItemsCategoriesView.ts
mattorchard, line 1: Can we do this part in a second PR?

Current approach is insufficient cause the subject terms also needs to split off of `;` and also tidy white-space, lowercase and stuff

Potentially still viable to do as a matview, but if we do it that way we should sanity check the perf is good

(also matviews not currently getting added to allDbTypes is a kysely limitation, so requires manual intervention if we do take that approach)

seeds/1746647286572_AddToTagCategories.ts
```diff
@@ -0,0 +1,14 @@
...
+  try {
+    const rawData = fs.readFileSync('seeds/categories_tags.json', 'utf-8');
+    const jsonData = JSON.parse(rawData);
+
+    await db.deleteFrom('TagCategories').execute();
+    await db.insertInto('TagCategories').values(jsonData).execute();
+    console.log('Seeded TagCategories table successfully');
+  } catch (error) {
+    console.error('Error reading or parsing JSON file:', error);
+  }
```
mattorchard, line 13: We prolyl don't want our own try/catch here since we want an error to be able to propagate to kysely so that the migration/seed is known to have failed

src/migrations/1746652473628_AddAgendaItemsCategoriesView.ts
melissaluc: Agreed. Can consider a separate second PR to address table restructuring and avoid joining on ANY(STRING_TO_ARRAY(...)). 

Action: Will remove from this PR.

seeds/1746647286572_AddToTagCategories.ts
melissaluc: Acknowledged, will remove b/c fs.readFileSync and JSON.parse will throw their own exceptions, allowing seed error to propagate to kysely ensuring failure is properly detected and logged to the stack trace.

melissaluc:
NOTE TO SELF: (Maybe)Fix rebase history before merging. Will investigate and update soon!

mattorchard (approved):

PR #175, opened 2025-05-30
Add OpenGraph image for site

17cliu:
Fixes #88

This PR sets an explicit OpenGraph image for the website, so that link previews (e.g. when you send a link to the site in Slack/WhatsApp/etc) now use the image we've selected for this purpose, instead of the first image found on the page.

**Before (used the image from the page header):**
image

**After (using our custom illustrations):**
 image

PR #176, opened 2025-05-30
Add error handling to Public Consultation page

17cliu:
While working on #175, the CI could not build the site -- it seemed to break on the public consultation page, because the fetched data was invalid JSON. However, the page loaded fine when run locally, so I assume that the API just didn't like the CI fetching from it.

This PR adds some error handling to the public consultation list so it doesn't crash if it fails to fetch data.

### DIFF
--- a/src/components/PublicConsultationList.tsx
+++ b/src/components/PublicConsultationList.tsx
@@ -6,14 +6,28 @@ const isUpcoming = (event: EventData) => {
 };
 
 export async function PublicConsultationList() {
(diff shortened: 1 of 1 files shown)


garobrik (approved):
LGTM! this page is currently not linked from anywhere, and we should have a proper ETL pipeline for it anyway instead of fetching whenever we build the site. first we need to figure out how to re-incorporate the info into the rest of the site in the first place!

Issue #147, opened 2025-05-14
feat: improve "Submit comment" and "Request to speak" button workflows


shrouxm:
### Description
Currently the "Submit comment" and "Request to speak" buttons open a `mailto` link with prefilled content. This isn't an ideal workflow because:
- `mailto` links don't work well everywhere
- we're severely restricted in how and what we can communicate to the user within the context of prefilled content in their email client

We'd instead like those buttons to open up a UI within Civic Dashboard, where we can display and collect structured information to help the user write the email, and then provide the option to send the resulting email via mailto link _or_ copy it to a clipboard. We can also then provide follow-up information upon the user taking the action.

PR #198, opened 2025-06-17
WIP: Add submit comment modal

17cliu:
Part of #147


17cliu:
The new modal is temporarily behind this new option in the dropdown menu:

image

Here's what the modal currently looks like. It has inputs for the user's name and comments, and as the user inputs this text, it gets prefilled in the email body to the right:

image

17cliu:
Per @garobrik's feedback on [Slack](https://civictechto.slack.com/archives/C06KU3DHEKV/p1750264848976209?thread_ts=1750138073.364359&cid=C06KU3DHEKV), here's what the latest iteration looks like:

image

The inputs are now "inlined" into the email body, which saves a lot of space and feels more natural. This also opened up the possibility of showing the long description for the comment field on the side instead of above/under it.

The layout is also responsive now -- though the description for the comment field ends up in what I would consider an awkward place.

jeromegv (approved):

Issue #125, opened 2025-04-30
Action buttons show on past agenda items


shrouxm:
**Describe the bug**
Agenda items that have happened in the past cannot be acted on anymore, but we still show call to actions on them.

**To Reproduce**

1. Go to actions page
2. Open advanced filters
3. Toggle "show past items"
4. Scroll down to agenda items that are in the past

**Expected behavior**
Should be no "Take Action" item.

**Screenshots**

![Image](https://github.com/user-attachments/assets/9672184d-754d-430e-a5a3-e10083762f1c)

**Context**
I believe `AgendaItemCard.tsx` would be a good place to make this change (hopefully it will just be a simple check of the current date against `agendaItem.meetingDate`.

younesabd100:
Hello I have managed to remove the option take action when the date of the meeting is anterior but it seems that the button " show past items " shows meeting in the future . Should i update it too ? 

![Image](https://github.com/user-attachments/assets/071e78da-c664-43ec-a226-1c87ef3896a3)

jeromegv:
fixed

PR #208, opened 2025-07-30
#125 dont show action buttons when meeting date is before today

jeromegv:
(take action, submit a comment, request) when meeting date is before today

### DIFF
--- a/src/components/AgendaItemCard.tsx
+++ b/src/components/AgendaItemCard.tsx
@@ -25,6 +25,7 @@ import Link from 'next/link';
 import { logAnalytics } from '@/api/analytics';
 import { AgendaItemCommentModal } from '@/components/AgendaItemCommentModal';
 
+const today = new Date();
 today.setHours(0, 0, 0, 0); // Reset to midnight
 
 const requestToSpeakHref = (item: AgendaItem, decisionBody: DecisionBody) => {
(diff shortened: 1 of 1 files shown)


src/components/AgendaItemCard.tsx
```diff
@@ -25,6 +25,9 @@ import Link from 'next/link';
 import { logAnalytics } from '@/api/analytics';
 import { AgendaItemCommentModal } from '@/components/AgendaItemCommentModal';
 
+const today = new Date();
+today.setHours(0, 0, 0, 0); // Reset to midnight
```
17cliu, line 29: This is a constant in the file, but outside of the component. So this date will just be defined once -- when the server is initialized. E.g. it'll always be July 31 until the server restarts. 

Since we're repeating the date comparison logic twice, I recommend putting this all into a helper function, e.g.:

```ts
function getIsDateIsUpcomingOrToday(dateString: string) {
  const today = new Date();
  today.setHours(0, 0, 0, 0); // Reset to midnight

  const date = new Date(dateString);
  date.setHours(0, 0, 0, 0);

  return date >= today;
}
```

17cliu (changes requested):
Requesting one change 🙏

src/components/AgendaItemCard.tsx
```diff
@@ -235,6 +251,9 @@ export function SearchResultAgendaItemCard({
   const {
     searchOptions: { textQuery },
   } = useSearch();
+  const itemDate = new Date(item.meetingDate);
+  itemDate.setHours(0, 0, 0, 0);
```
17cliu, line 255: These lines are unneeded now - please delete!

17cliu (approved):
one comment -- otherwise LGTM! 🚀

src/components/AgendaItemCard.tsx
jeromegv: thats what happens when i rush a commit before end of hacknight lol

Issue #211, opened 2025-08-06
Add seed command to migration script


melissaluc:
### Description

Additional steps to run after migrations to setup the database

### To-do (Acceptance Criteria)

- [x] Updates the database schema
- [x] Test to see if it works on another machine

jeromegv:
@melissaluc Can you confirm if there was anything else pending?

melissaluc:
@jeromegv nothing left pending, good to close!

PR #213, opened 2025-08-06
Added seed command to migration script

melissaluc:
## Description

- package.json was changed to script "db:run-migrations"

## Related issue

#211 

### Common tasks:

- [ ] All to-do items (acceptance criteria) from ticket have been treated

### DIFF
--- a/package.json
+++ b/package.json
@@ -16,7 +16,7 @@
     "db:update-types": "kysely-codegen --out-file ./src/database/generatedDbTypes.d.ts --url $DATABASE_URL && eslint ./src/database/generatedDbTypes.d.ts --fix",
     "db:create-migration": "kysely migrate:make",
     "db:list-migrations": "kysely migrate:list",
-    "db:run-migrations": "kysely migrate:latest",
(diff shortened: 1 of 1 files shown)


melissaluc:
@jeromegv just need to see if there are no issues on someone else's machine

jeromegv:
ok will test in my local env and let you know!

package.json
```diff
@@ -16,7 +16,7 @@
     "db:update-types": "kysely-codegen --out-file ./src/database/generatedDbTypes.d.ts --url $DATABASE_URL && eslint ./src/database/generatedDbTypes.d.ts --fix",
     "db:create-migration": "kysely migrate:make",
     "db:list-migrations": "kysely migrate:list",
-    "db:run-migrations": "kysely migrate:latest",
+    "db:run-migrations": "kysely migrate:latest && kysely seed:run",
     "db:run-seed": "kysely seed:run",
```
17cliu, line 20: looks like we already have a `run-seed` script here :O so an alternative to this PR could be to just list this command after the `run-migrations` in the readme.

Issue #58, opened 2025-03-12
feat: filter/sort options on actions page


shrouxm:
It would be nice to allow the user to filter/sort the actions page search, at the very least on relevance and meeting date!

PR #205, opened 2025-07-25
WIP: Add request to speak modal

17cliu:
Part of #147

PR #218, opened 2025-08-27
Filter or sort options issue 58

KarishmaChanglani:
This is to add a sorting drop down using the current sorting filters, I have not added any new search options and I have kept the design similar to the decision body selection dropdown. I have updated the current combo box to be useful for drop downs that don't require a search bar.

### DIFF
--- a/src/components/search.tsx
+++ b/src/components/search.tsx
@@ -38,7 +38,7 @@ export function SortDropdown() {
 
   const onSelect = useCallback(
     (selectedId: number) => {
-      const selectedOption = Object.values(sortByFilterOptions).find(
+      const selectedOption = sortByFilterOptions.find(
         (opt) => opt.sortId === selectedId,
       );
       setSearchOptions((opts) => ({


KarishmaChanglani:
> Great job! Very useful to be able to check on old items and sort in reverse date
> 
> 1. Not really a code issue but more of a UI/UX opinion, I don't fully see the point of showing "least relevant". We know it's going to be bad, and if we want to keep the options relevant on how a typical user would use the tool, i'd suggest to remove it.
> 2. Also from a UI standpoint, i would auto-select the current sorting being used. When it's blank, we can't really tell what is the sorting currently being used on the page we are looking at (without changing back and forth)
> 
> Adding a screenshot to clarify what I mean Screenshot 2025-09-01 at 10 20 04 PM

All good points! I will make those updates with "Most Relevant" as the default (since I believe that is the current sorting without the dropdown)

KarishmaChanglani:
> I have a few comments I'd like addressed before approving 🙏

Thanks for the comments, I will update soon, I apologize for the late response.

17cliu:
(merged in main branch so we can get the latest CI fixes)

src/components/search.tsx
```diff
@@ -14,10 +14,66 @@ import {
...
+export function SortDropdown() {
+  const {
+    searchOptions: { sortBy, sortDirection },
+    setSearchOptions,
+  } = useSearch();
+
+  // Use sortByFilterOptions instead of sortByOptions
+  const options = useMemo(
+    () =>
+      Object.values(sortByFilterOptions).map((opt) => ({
```
17cliu, line 32: since `sortByFilterOptions` is now an array, you can just use `map` on it directly:

```suggestion
      sortByFilterOptions.map((opt) => ({
```

src/components/search.tsx
```diff
@@ -14,10 +14,66 @@ import {
...
+        id: opt.sortId as number,
+        label: opt.sortLabel as 'Oldest' | 'Newest' | 'Most Relevant',
+      })),
+    [],
+  );
+
+  const onSelect = useCallback(
+    (selectedId: number) => {
+      const selectedOption = Object.values(sortByFilterOptions).find(
```
17cliu, line 41: same here:

```suggestion
      const selectedOption = sortByFilterOptions.find(
```

src/components/search.tsx
```diff
@@ -14,10 +14,66 @@ import {
...
+          | 'descending'
+          | undefined,
+      }));
+    },
+    [setSearchOptions],
+  );
+
+  // Find the selected option's id based on sortBy value
+  const selectedId = useMemo(() => {
+    const selectedOption = Object.values(sortByFilterOptions).find(
```
17cliu, line 58: here too:

```suggestion
    const selectedOption = sortByFilterOptions.find(
```

17cliu (changes requested):
thank you for making those changes! just a bit more cleanup and i think we're ready 🙏 

--- 

side note: for referencing the issue, it's better to add it in the issue description instead of the issue title. you can use `Resolves #58` or `Fixes #58`, which will cause github to automatically close the issue when this PR is merged. you can read more about this github feature [here](https://docs.github.com/en/get-started/writing-on-github/working-with-advanced-formatting/using-keywords-in-issues-and-pull-requests), and here's an [example PR](https://github.com/civic-dashboard/civic-dashboard-web/pull/205) that used this.

PR #220, opened 2025-09-02
Reduce image sizes

17cliu:
The images we use in our OpenGraph data and on the How Council Works page are all fairly large PNGs (~500kb each). 

This PR replaces those images with JPG versions that are nearly 90% smaller (~50kb each). I compressed the images using https://squoosh.app, with MozJPEG at 85% quality.

This PR also reduces the file size for the PNG logo by 30% (125kb -> 88kb). I compressed it using OxiPNG at effort level 6.


17cliu:
> Do you want to change logo.png for logo.jpg while we are at it? @17cliu

No, because the logo needs to have a transparent background. If we had a svg of the logo, we could use that instead. But the logo png is not too big, so I'm not stressed about it.

jeromegv:
Good point about transparency! Looks good to me

sarmilan (approved):

PR #223, opened 2025-09-07
Added pagination to councillor's page for motions - Pagination for Councillors Page issue #224

pliakhov:


pliakhov:
> @pliakhov Would you mind fixing your lint (you can use `eslint`), this will auto-generate a preview URL once the lint passes, which should facilitate the review

Ok will do!

PR #229, opened 2025-09-15
Pliakhov/pagination

pliakhov:


src/app/councillors/[contactSlug]/components/AgendaItemResults.tsx
```diff
@@ -42,6 +41,156 @@ const AgendaItemCard = memo(function AgendaItemCard({
...
+
+    return Array.from({ length: end - start + 1 }, (_, i) => start + i);
+  };
+
+  const visiblePages = getVisiblePages();
+
+  if (totalPages <= 1) return null;
+
+  return (
+    <nav className="flex items-center justify-center space-x-2 mt-8">
```
Comment (17cliu) [inline src/app/councillors/[contactSlug]/components/AgendaItemResults.tsx:137]: this needs some bottom margin:

image

```suggestion
    <nav className="flex items-center justify-center space-x-2 mt-8 mb-8">
```

17cliu (approved):
thanks for making those changes! requesting one more small fix, and this is good to go 👍  i don't think this PR is causing the cloudflare issue, although maybe it's worth waiting a little bit to see if @jeromegv can investigate.

---

one thing i realized is that we forgot to check how this looks on mobile! 😅 

image

the prev/next buttons can be cut off. we don't need to hold up this PR for this though -- please tackle this in a follow-up PR!

jeromegv:
oh you meant accessing the councillor page on the preview link is broken, interesting, ok will check when have a second

jeromegv:
So we don't have access to the preview URLs however my guess based on a short investigation on slack is that we hit CPU limits when calling the councillor page (the same way we hit those limits in prod)
https://civictechto.slack.com/archives/C07DVPT2C0J/p1758249794104329

I'm looking at the PR and wondering, are we actually reducing the size of the DB call to only get the items in the councillor page that we need on this page or are we still hitting ALL items but just implementing UI pagination after we've received 100% of the items? 

Because that would explain why we still hit the limits even with pagination implemented.

pliakhov:
Thank you! @17cliu should I merge the pr then?

Issue #56, opened 2025-03-12
feat: return list of tags on agenda items when querying for agenda items


shrouxm:
We've already implemented the logic to query for all agenda items matching a given tag, but we haven't implemented the logic to, given an agenda item, display which tags it matches. This issue covers the backend work to return those tags in `app/api/agenda-item/search/route.ts`, and the to-be-created route for fetching a single agenda item by reference.

The existing related code is quite messy, so feel free to ask for more info. The function `queryAndTagsToPostgresTextSearchQuery` could be a good place to start.

PR #234, opened 2025-09-17
Issue 56 agenda item tags

Andre053:
Added feature to show related tags of agenda items as requested by [Issue #56](https://github.com/civic-dashboard/civic-dashboard-web/issues/56). I have added a comment on the issue to clarify if this is what the issue was meant to address, or whether additional tag functionality on the item list and search was desired.

This solution is isolated to the agenda item card. When a card is clicked, matching tags are shown at the bottom of the card. Each tag is a link that redirects back to the item list with the clicked tag already selected. 

The redirection is achieved using URL parameters to identify the tag selected in the form ?tag="<tag name>". There is additional functionality to remove the parameters in the URL, though it may be best to leave them. I did not see any other uses of URL parameters within the project and this solution seems hacky, should the handling of URL parameters be generalized and separated from the current file?

A fleeting issue was occurring during testing when the /actions endpoint with manually entered parameters was attempted to be accessed, though on subsequent tests it has not occurred. I believe the problem was with the search context being accessed before it was set, I reordered the logic and have not seen the issue again.

### DIFF
--- a/src/components/AgendaItemList.tsx
+++ b/src/components/AgendaItemList.tsx
@@ -109,36 +109,36 @@ function AgendaItemListInner({ initialSearchParams }: Props) {
 
   return (
     <div className="flex flex-col space-y-4 p-4 items-stretch max-w-full sm:max-w-max-content-width">
-        <div className="mt-4 mb-2">
-          <h1 className="text-2xl font-bold">Actions</h1>
-          <p>
-            Here are agenda items that the City of Toronto will discuss at
-            upcoming meetings. You can provide feedback on these items by
-            submitting comments by email, which will be read at the meeting, or
-            requesting to speak at the meeting live, in person or over video
-            conferencing.
-          </p>
-        </div>
-        <hr />
-        <div className="flex flex-row self-stretch items-center space-x-2">
-          <div className="flex-grow">
-            <SearchBar />
-          </div>
-          <div className="sm:max-w-max-content-width">
-            <SortDropdown />
-          </div>
-        </div>
-        <Tags />
-        <hr />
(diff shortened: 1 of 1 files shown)


17cliu (changes requested):
very cool, and very useful! it'll be great to be able to link to other related actions 👍 

i do have a suggested change, explained further below, with the suggested code changes supplied in this extra PR: #254. it's set to merge into this PR, so if you approve it, you can roll the changes into this PR that way.

17cliu:
i'm looking at the preview link and the layout seems broken 🤔 

image

17cliu:
also, clicking on the tags doesn't update the url with the selected tag...

17cliu (changes requested):
i'm pretty sure the issue is that there's an extra `<SearchProvider>` wrapper, which slipped back in during the merge conflict resolution. please remove it from `AgendaListInner` component and hopefully that should fix it...

src/components/AgendaItemList.tsx
```diff
@@ -108,39 +107,19 @@ function AgendaItemListInner({ initialSearchParams }: Props) {
...
-          <h1 className="text-2xl font-bold">Actions</h1>
-          <p>
-            Here are agenda items that the City of Toronto will discuss at
-            upcoming meetings. You can provide feedback on these items by
-            submitting comments by email, which will be read at the meeting, or
-            requesting to speak at the meeting live, in person or over video
-            conferencing.
-          </p>
-        </div>
-        <hr />
```
17cliu: oops, we lost this intro section :')

Issue #238, opened 2025-09-24
Implement backend pagination when loading councillor page


jeromegv:
**Describe the bug**
Go to the councillor page. Click any councillor. There's a high chance you might crash the website. 

We are getting reports from cloud flare that it's using too much CPU cycle. 

The issue is that every time we load that page we query the entire councillor voting record, which gets larger and larger every month. 

We need pagination at the backend level to only load what's needed on the first page. 

Seems to happen for almost any councillor page, if it doesn't happen, if you refresh a few times it will surely happen

Issue #261, opened 2025-10-13
If the URL for a councillor a page is wrong, should return 404 instead of 500 (SEO related)


jeromegv:
### Description

Try this URL (which is a former councillor)
https://civicdashboard.ca/councillors/jennifer-mckelvie
It returns: 500 Internal Server Error.
Same if you try any URL after /councillors/ that is not found

Instead of 500, it should return 404 not found, it's better for SEO and it's also better for our app not to be throwing that kind of errors when something is not found.

I had fixed a similar bug for the action page few months ago, so the fix should be similar:
https://github.com/civic-dashboard/civic-dashboard-web/issues/202

### To-do (Acceptance Criteria)

- Go to: https://civicdashboard.ca/councillors/jennifer-mckelvie
- Should return 404 not found

PR #265, opened 2025-10-14
Added backend pagination - Issue #238

Andre053:
This commit solves issue #238. 

The solution involves using URL parameters to keep track of the current page. When the page is changed, with the frontend pagination code, it triggers a backend request to get the data for that specific page. 

A default of 10 items per page is set, and therefore 10 items will be returned from each backend request. The backend request contains a header that states the total number of items for the specific councillor. This item count is used for the frontend pagination to be able to show the correct amount of total results and the corresponding amount of pages.

The backend requests for getting agenda items were extracted into an API route. The original agenda item request was changed to return the 10 most recently updated items, offset by the product of the page number received and the page size. This is also where a new query was created to get the count of all agenda items for the councillor. 

The initial request is parsed on the first page load, if no page parameter is found, it defaults to page 1. 

NOTE: This change breaks the frontend search functionality. While code was added to filter the agenda items on the page in accordance with the search term, the search will only ever execute on the 10 items on the page. Should we remove the search functionality until we decide to implement this properly?

### DIFF
--- a/src/logic/councillorItems.ts
+++ b/src/logic/councillorItems.ts
@@ -40,17 +40,9 @@ async function getTotalAgendaItemsForContact(
 ): Promise<{ itemCount: number }[]> {
   const agendaItemCount = db
     .selectFrom('Votes')
-    .innerJoin('Motions', (eb) =>
-      eb
-        .onRef('Votes.agendaItemNumber', '=', 'Motions.agendaItemNumber')
-        .onRef('Votes.motionId', '=', 'Motions.motionId'),
-    )
-    .innerJoin('AgendaItems', (eb) =>
-      eb.onRef('Votes.agendaItemNumber', '=', 'AgendaItems.agendaItemNumber'),
-    )
     .select((eb) => [
       eb.fn
-        .count<number>('AgendaItems.agendaItemNumber')
+        .count<number>('Votes.agendaItemNumber')
         .filterWhere('Votes.contactSlug', '=', contactSlug)
         .distinct()
         .as('itemCount'),
--- a/src/app/councillors/[contactSlug]/components/AgendaItemResults.tsx
+++ b/src/app/councillors/[contactSlug]/components/AgendaItemResults.tsx
@@ -203,11 +203,11 @@ const Pagination: React.FC<PaginationProps> = ({
 export default function AgendaItemResults({
   contactSlug,
   currentPage,
-  searchTerm = '',
+  //searchTerm = '',
 }: {
   contactSlug: string;
   currentPage: number;
-  searchTerm?: string;
+  //searchTerm?: string;
 }) {
   if (currentPage <= 0) currentPage = 1;
 
@@ -216,8 +216,10 @@ export default function AgendaItemResults({
   const [pageAgendaItems, setPageAgendaItems] = useState<AgendaItem[]>([]);
   const [agendaItemCount, setAgendaItemCount] = useState(0);
 
+  /*
   const tidySearchQuery = searchTerm.toLocaleLowerCase().trim();
 
+  // if frontend search is re-enabled, use filteredItems in the JSX result instead of pageAgendaItems
   const filteredItems = useMemo(
     () =>
       tidySearchQuery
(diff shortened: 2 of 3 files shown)


jeromegv:
`NOTE: This change breaks the frontend search functionality. While code was added to filter the agenda items on the page in accordance with the search term, the search will only ever execute on the 10 items on the page. Should we remove the search functionality until we decide to implement this properly?`

Yes, we can create a ticket to re-add it as a separate work effort, but i agree we shouldn't just search within 10 records

Andre053:
Yes, I'm seeing it is crashing immediately on selecting a councillor, seemingly due to a backend issue. I will try and investigate why this is happening on the preview

src/app/councillors/[contactSlug]/page.tsx
```diff
@@ -39,108 +39,43 @@ async function getCouncillor(db: Kysely<DB>, contactSlug: string) {
...
-      'OriginalSummaries.agendaItemSummary',
-      sql<string>`CONCAT("Motions"."yesVotes", '-', "Motions"."noVotes")`.as(
-        'tally',
-      ),
-      'aiSummary',
-    ])
-    .execute();
+async function getAgendaItems(contactSlug: string) {
+  const response = await fetch(
+    `http://localhost:3000//api/councillor-items?contactSlug=${contactSlug}&page=1&pageSize=10`,
```
Comment (jeromegv) [inline src/app/councillors/[contactSlug]/page.tsx]: I think this might be the issue, i don't think we can call localhost:3000 on a cloudflare worker instance @Andre053 
That's typically not how we call ourself

Andre053:
Still receiving issues when trying to get the app URL using next/headers on server side

garobrik:
@Andre053 i might be able to help debug this! could you share the version of the code where you're calling with the same format as `src/logic/search.ts`? i.e. in your case i think it would look like ``const response = await fetch(`/api/councillor-items?contactSlug=${contactSlug}&page=${currentPage}&pageSize=10);``

jeromegv:
I've checkout your branch in my local environment but the councillor page is also not working, so it doesn't seem to be an issue with cloudflare right now (at least for now)
Screenshot 2025-11-15 at 10 14 43 AM

@Andre053

Andre053:
Got a working deployment! Changed the logic based on @jeromegv's advice and that seems to have fixed it -- thanks for looking into this.

The functionality is there now, but I think I might be able to optimize it a bit more. I can also move the logic code into src/database, currently it is in src/logic

jeromegv:
Awesome! Haven't reviewed the code yet but tested it with the testing link and it loads properly. Refreshing multiple times also doesn't crash the website, so big wins!

I think we should still go forward with removing the search bar, as it would be confusing that it only search in the results being shown, and have a separate github issue to re-add search in the future.

Will let you know once i've spent a bit more time looking at the code

src/logic/councillorItems.ts
```diff
@@ -0,0 +1,188 @@
...
+    page,
+    pageSize,
+  );
+  const itemCountObj = await getTotalAgendaItemsForContact(db, contactSlug);
+  const itemCount = itemCountObj[0]['itemCount'];
+
+  return [agendaItemsResult, itemCount];
+}
+
+async function getTotalAgendaItemsForContact(
```
jeromegv: I think the SQL could be simplified here to improve performance, because we don't actually need to do all those joins. We just need the total number of votes from this councillor.
I just tested this query direclty against the db
```
SELECT COUNT(DISTINCT "agendaItemNumber") AS "itemCount"
FROM "Votes"
WHERE "contactSlug" = 'amber-morley';
```
and it was the same result as what i see here:
https://d8cd183c-civic-dashboard-web.civic-dashboard.workers.dev/councillors/amber-morley

jeromegv:
Putting the wrong councillor name seems to crash the app:
https://d8cd183c-civic-dashboard-web.civic-dashboard.workers.dev/councillors/alejandra-bravo2

This is what it shows in localhost
Screenshot 2025-11-23 at 10 16 17 PM

I realized it was an issue i had documented in the past with #261 
so i went ahead and pushed a fix for this, as this wasn't due to your code, but existing bug in the code base, which will be fixed with this PR too

Issue #221, opened 2025-09-03
Implement a markdown to wiki technology


jeromegv:
### Description

The idea would be to host the different wiki pages on our Github repository. Each wiki page would be written in markdown

Volunteers could edit the  markdown directly in github, which would automatically update a static page on the website under https://civicdashboard.ca/wiki

We need to do some research as there might be some different tools doing something similar

jeromegv:
Some suggestions:

https://nextjs.org/learn/pages-router/dynamic-routes-render-markdown

11ty is also a common tool often used in similar use casees (mentioned by @17cliu )
https://www.11ty.dev

PR #268, opened 2025-10-19
Markdown to HTML feature for automating .md files into Civic Dashboard Wiki

pliakhov:
This is resolving issue #221 markdown to wiki technology, for creating wiki documents, create .md wiki files in the contents/markdown folder and they will be created as wiki docs under the wiki tab in the dashboard website


jeromegv:
@pliakhov You are getting closer!

1. You are generating HTML files as part of the pipeline, so that works!
Your HTML files are here:
https://ad521471-civic-dashboard-web.civic-dashboard.workers.dev/html/
And they would get updated every single time we update any of those files on github, so that's great.

2.
However the issue with page.tsx is that it uses `fs`, i can't find where I wrote those instructions last week but `fs` doesn't actually work with cloudflare, cloudflare has no concept of file system, so it must be avoided entirely.  
Instead you must use `fetch` to query the html file, which will work with cloudflare. 
And in order to know what is in the directory to have the list of files in the wiki, since you do not have `fs`, you can have it generate a manifest.json file at build time, and have your page.tsx file read from that manifest to know all the html files generated

jeromegv (approved):
APPROVED

jeromegv:
@pliakhov Amazing, MERGE THAT IN!

PR #270, opened 2025-10-24
ParseQuesry Documentation

gurpreet11307:
Added the document PARSEQUERY.md and embedded comments in parseQuery.ts


jeromegv:
Hi @gurpreet11307 , this is good, however your lint for the comments you added is failing, which is the process that ensure we keep a consistent way of formatting our code.
Is that something you think you can fix or you want me to fix it? We use `eslint` for our project, i have the `eslint` extension in Visual Studio Code to help me with that. But I understand if you want us to fix it on your behalf!

PR #331, opened 2026-01-17
Fix styling for Join page (and start building out design system components)

17cliu:
## Description

Resolves #288

This PR fixes the "Join" page looking visually different from our other text pages. For example, here was the Join page vs About page before this PR:

| Join |  About |
| -- | -- |
| image |  image |

Here is what the Join page looks after: 

| Before | After |
| -- | -- |
| image |  image |

To ensure that we start reducing this problem of having slightly different styles everywhere, I created reusable components for page and section wrappers, heading levels 1 and 2, and bulleted lists. I've only started using them for the Join page, to keep this PR small, but the intention is to start converting such elements on other pages to these components. The styles I've used for these reusable components are still up for tweaking, of course -- this is just what I've started with based on what I've seen on the Join and About pages.

## Testing instructions

- Browse to `/join` on the preview site 
- Verify that the page looks good! It should no longer have the odd white background in dark mode. Also, check the page...
  - at various viewport widths 
  - on dark vs light mode (usually a browser or operating system setting)

## Checklist

- [x] This PR addresses all requirements described in the issue
- [x] I have commented my code, particularly in hard-to-understand areas
- [x] I have performed a self-review of my code
- [x] I have tested these changes in the preview deployment
- [x] I have made corresponding changes to the documentation (if relevant)

Reminder to self:
- [x] file issues to update other pages to use components after this PR is merged

PR #339, opened 2026-01-23
Update "How Council Works" page to use design system components

17cliu:
## Description

This PR updates the How Council Works page so that it uses the new reusable components introduced in #331 for things like headings, lists, etc, so it is more consistent with the rest of the site.

It also replaces the double `<br>` tags that separate paragraphs, with `<p>` tags around each paragraph.

Most obvious visual change you'll notice is that the headings are now left-aligned instead of centered, and use the same container width as the rest of the page text/images:

| Before | After |
|-|-|
| image | image

This PR introduces **2** new components, `Heading3` and `ArticlePage`, which is a narrower container than `Page`. This page looked _really_ weird using `Page`, which is a very wide container. It's awkward to read paragraphs that span so far across the screen, and also the images were either (a) too big if left to stretch to fill the container width, or (b) awkwardly positioned if they had a smaller max width, no matter if I left-aligned or centered them. You can see examples of these alternatives in the expandable sections below.

<details>
<summary>
Example showing this page using the existing wide `Page` container, with full-width images
</summary>
image
</details>

<details>
<summary>
Example showing this page using the existing wide `Page` container, with centered images
</summary>
image
</details>

## Testing instructions

- Open the `/how-council-works` page on the preview site.
- Verify it looks good.

## Checklist

- [x] I have tested these changes in the preview deployment

### DIFF
--- a/src/app/about/page.tsx
+++ b/src/app/about/page.tsx
@@ -1,5 +1,8 @@
 import { Metadata } from 'next';
 import { ExternalLink } from '@/components/ExternalLink';
+import { Page } from '@/components/ui/page';
+import { Section } from '@/components/ui/section';
+import { Heading1, Heading2, BulletedList } from '@/components/ui/text-items';
 
 export const metadata: Metadata = {
   title: 'About Us – Civic Dashboard',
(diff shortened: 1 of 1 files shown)


17cliu:
would appreciate thoughts on the page width/image positioning thing! (will share this on the ux channel for feedback too)

PR #393, opened 2026-04-03
Synchronize councillor voting UI with action search results

### LINKED ISSUE #151: Councillor vote individual agenda item page
### Description

This is about adding a way to see the individual item page, from the councillor page (like we do on action when we click on the item/ click learn more

jeromegv:
## Description
Resolves #151 

This bring the UI of the action search result page to the councillor page. This isn't meant to be a full UI revamp or a real improvement, it's just to make it more consistent between both AND allow the link to the "Learn more" (individual agenda item) which was missing entirely on the councillor page.

- Update councillor agenda item cards to use the standard Card-based UI for visual consistency with the main actions/search results page.
- Link to the "Learn more" individual agenda item page that we already had on the Actions page
- Implement a unified content overflow mechanism with a gradient fade, replacing the manual "Show more" logic in SummaryPanel.
- Modify councillorItems.ts logic to use innerJoin for city summaries, ensuring voting records are only displayed when official TMMIS summary data is available.
- Include itemStatus in the councillor data pipeline and display it using the sentenceCase utility.

Before:
Screenshot 2026-04-02 at 10 02 19 PM

After:
Screenshot 2026-04-02 at 10 27 13 PM

## Testing instructions

Go on the Councillor page, click a councillor, click around to ensure everything works.

## Checklist

- [X ] This PR addresses all requirements described in the issue
- [ ] I have tested these changes in the preview deployment

### DIFF
--- a/src/app/councillors/[contactSlug]/components/AgendaItemResults.tsx
+++ b/src/app/councillors/[contactSlug]/components/AgendaItemResults.tsx
@@ -66,7 +66,7 @@ const AgendaItemCard = memo(function AgendaItemCard({
           <span className="font-bold">{firstMotion.committeeName}</span>
         </CardContent>
 
-        <CardContent className="flex-grow">
+        <CardContent className="flex-grow pb-0">
           <div className="relative max-h-[250px] overflow-hidden">
             <div
               className="absolute inset-0 h-[100px] top-[150px] bg-gradient-to-t from-white dark:from-neutral-800 from-1% via-transparent to-transparent pointer-events-none"
--- a/src/app/councillors/[contactSlug]/components/AgendaItemResults.tsx
+++ b/src/app/councillors/[contactSlug]/components/AgendaItemResults.tsx
@@ -8,13 +8,7 @@ import { MotionsList } from '@/app/councillors/[contactSlug]/components/MotionsL
 import { Button, buttonVariants } from '@/components/ui/button';
 import { useRouter } from 'next/navigation';
 import { useSearchParams } from 'next/navigation';
-import {
-  Card,
-  CardContent,
-  CardFooter,
-  CardHeader,
-  CardTitle,
-} from '@/components/ui/card';
+import { Card, CardContent, CardHeader, CardTitle } from '@/components/ui/card';
 import { Chip } from '@/components/ui/chip';
 import Link from 'next/link';
 import { cn } from '@/components/ui/utils';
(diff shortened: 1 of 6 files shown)


17cliu:
yay for more consistency! 👏 

one thing that is unclear: how do we see the rest of this text? in the current ui on prod, there's a clear "show more" link, which expands the card. i don't think it would be obvious to users unfamiliar with civic dashboard that they should click the reference number in the top right corner to see the rest of the text in a new tab. one option would be to add a show more link where the text fades, and that link would also just open the action item in a new tab.

image

edit: i just saw the "learn more" button now 😩 the card was so big that i didn't see it until i scrolled. i think it would still be better to put that functionality where the text fading/truncation happens!

17cliu:
another thing to think about with the non-expandable truncation: screenreaders. since the text is all still there, it believe it will read it all. i don't think we need to handle this immediately in this PR, but something to consider. i think the easiest way is to truncate the text before rendering -- otherwise, we would have to detect what text is visually rendered, and then mark the rest as not visible... complicated! 

if we mark-up the big summary text section correctly though, a screenreader user should be able to just skip out of it and into the next thing... so maybe that's the easiest solution.

17cliu:
tangent: we got a keysmash status here 😂 

image

jeromegv:
> tangent: we got a keysmash status here 😂
> 
> image


For this, I created this other issue:
https://github.com/civic-dashboard/civic-dashboard-web/issues/402

jeromegv:
I moved LEARN MORE button just where the text is on the councillor page
Screenshot 2026-04-08 at 11 55 19 PM


And then ended up doing the same for the Actions page, it's optional, but curious to hear your thoughts on it. 
Screenshot 2026-04-08 at 11 56 27 PM

Let me know if it's good to merge @17cliu

17cliu (approved):
i like the "learn more" button on the councillors page a lot! 

however, i prefer the old style for the actions page, for a few reasons:

- it's weird that when i hover over the card, the card and the "learn more" button change background at different rates
- the empty space introduced by the row that the "learn more" button adds + the increased empty space in the footer right below it (now that there's only one button) looks really odd

so i'd like to see the changes for the actions page reverted. but approving so you can merge immediately when you're ready!

jeromegv:
i agree with you, wanted to see your thoughts but yeah it doesn't look as great on the action page.

PR #403, opened 2026-04-10
Homepage update styles phase

antler-hat:
## Description
This is a series of almost entirely style-related changes. The purpose is to consolidate, systemize and update styles in accordance with our evolving design system, to ensure consistency and scaleability in advance of homepage updates

Note that the PR is into a feature branch instead of `main`, and I'll be making subsequent into this feature branch

## Before and After
screencapture-civicdashboard-ca-2026-04-19-20_51_57
screencapture-c71ff9a1-civic-dashboard-web-civic-dashboard-workers-dev-2026-04-19-20_54_07

## Checklist

- [n/a] I have commented my code, particularly in hard-to-understand areas
- [x] I have tested these changes in the preview deployment
- [n/a] I have made corresponding changes to the documentation (if relevant)

### DIFF
--- a/src/app/coming-soon/page.tsx
+++ b/src/app/coming-soon/page.tsx
@@ -11,11 +11,11 @@ export default function ComingSoon() {
       <main className="flex-grow flex items-center justify-center bg-gradient-to-br from-gray-50 to-gray-100 dark:from-gray-900 dark:to-gray-800">
         <div className="max-w-3xl mx-auto px-4 py-16 text-center">
           <div className="space-y-8">
-            <h1 className="text-4xl sm:text-5xl md:text-6xl font-bold tracking-tight text-transparent bg-clip-text bg-gradient-to-r from-blue-600 to-purple-600 dark:from-blue-400 dark:to-purple-400 pb-2">
+            <h1 className="text-transparent bg-clip-text bg-gradient-to-r from-blue-600 to-purple-600 dark:from-blue-400 dark:to-purple-400 pb-2">
               Coming Soon
             </h1>
 
-            <p className="text-xl sm:text-2xl text-gray-600 dark:text-gray-300">
+            <p className="text-gray-600 dark:text-gray-300">
               Our friendly neighbourhood volunteers are hard at work building
               this page. Please check back soon!
             </p>
@@ -26,7 +26,7 @@ export default function ComingSoon() {
 
             <Link
               href="/"
-              className="inline-block px-8 py-4 bg-gradient-to-r from-blue-600 to-purple-600 text-white rounded-full font-medium hover:from-blue-700 hover:to-purple-700 transition-all transform hover:scale-105 shadow-lg"
+              className="text-button inline-block px-8 py-4 bg-gradient-to-r from-blue-600 to-purple-600 text-white rounded-full hover:from-blue-700 hover:to-purple-700 transition-all transform hover:scale-105 shadow-lg"
             >
               Back to Home
             </Link>
--- a/src/app/globals.css
+++ b/src/app/globals.css
@@ -5,9 +5,8 @@
 :root {
   --background: #ffffff;
   --foreground: #171717;
-  --font-body: var(--font-geist-sans);
-  --font-heading: var(--font-geist-sans);
-  --font-mono: var(--font-geist-mono);
+  --font-body: var(--font-ibm-plex-sans);
+  --font-heading: var(--font-ibm-plex-sans);
 }
 
 @media (prefers-color-scheme: dark) {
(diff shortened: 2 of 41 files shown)


src/app/coming-soon/page.tsx
```diff
@@ -11,11 +11,11 @@ export default function ComingSoon() {
       <main className="flex-grow flex items-center justify-center bg-gradient-to-br from-gray-50 to-gray-100 dark:from-gray-900 dark:to-gray-800">
         <div className="max-w-3xl mx-auto px-4 py-16 text-center">
           <div className="space-y-8">
-            <h1 className="text-4xl sm:text-5xl md:text-6xl font-bold tracking-tight text-transparent bg-clip-text bg-gradient-to-r from-blue-600 to-purple-600 dark:from-blue-400 dark:to-purple-400 pb-2">
```
17cliu, line 14: comment: huh, i didn't even know we had [this page](https://c71ff9a1-civic-dashboard-web.civic-dashboard.workers.dev/coming-soon)!

src/app/globals.css
```diff
@@ -48,7 +110,11 @@ html {
 body {
   color: var(--foreground);
   background: var(--background);
-  font-family: Arial, Helvetica, sans-serif;
+  font-family: var(--font-body), sans-serif;
+  font-size: 1rem;
+  font-weight: 400;
+  line-height: 1.5;
+  letter-spacing: 0;
```
17cliu, line 117: i would recommend not setting default styles on `body` -- by adding them here, these styles apply to everything, which means when you _don't_ want these styles, you have to forcibly override them. the approach we've been using throughout this project so far has been to apply styles to each element as needed, and adding styles to `body` changes that.

src/app/globals.css
```diff
@@ -1,39 +1,101 @@
...
+    font-family: var(--font-body), sans-serif;
+    @apply text-[1rem] font-normal leading-[1.5];
   }
 
-  :root {
-    --radius: 0.5rem;
+  .text-small {
+    font-family: var(--font-body), sans-serif;
+    @apply text-[0.875rem] font-normal leading-[1.5];
+  }
```
17cliu, line 59: so, this is a totally valid approach to do theming (write a css class and use it on the element) -- and actually this is how i often do styling on my own personal projects.

however, this project uses tailwind (a css utility library), where styling is usually NOT by writing custom css rules, but by applying a group of tailwind css classes to elements (as you've seen on all the components). so, introducing the new custom styling classes creates a divergence from this approach. i worry that:

1. other devs in the future (maybe when we are not around to share the project history with them) will forget to use these custom classes and just use tailwind classes, because they see it's a tailwind project, and don't realize there are custom classes; and
2. it'll be hard to identify which classes are custom and which are tailwind, because they look similar. this makes it harder to hunt down the source of a particular style.

looking at the contents of `.text-small`, as an example, all the styles aside from font-family can be set as `text-sm leading-6` (or alternatively, `text-sm/[1.5]` -- i also find it hard to decipher what the [tailwind leading values](https://tailwindcss.com/docs/line-height) mean).

for the font-family, it looks like we can introduce custom `font-*` css classes into the tailwind system, by registering them into `tailwind.config.ts`. so we can add `font-body` and `font-heading` by adding something like this:

```js
const config: Config = {
  // ...truncated
  theme: {
    extend: {
      // ...truncated
      fontFamily: {
        'heading': ['IBM Plex Sans', 'sans-serif'],
        'body': ['Epilogue', 'sans-serif'],
      }
    },
  },
  // ...truncated
};

**Review by 17cliu — COMMENTED** [review]
thanks for taking this on! this is a good start -- i really like the new font! 

two main things:
1. in general, PRs should be small and focused. that makes them easier to review (easier for the reviewer to conceptualize what's happening, and to catch bugs), and if anything goes wrong, it's easier to fix (we can revert just a targeted set of changes instead of everything). in the case of styling changes like this, the styling changes for each page can be a separate PR. 
2. i'm not sure that we should be introducing many custom css classes in a tailwind system -- more on that in the comment below. 

so what i would like to see is that we start pulling these changes out into separate PRs, one at a time, and then getting them reviewed and merged. to begin with, i can help with getting the fonts set up. i'll ping you on that PR when it's up.

File: src/app/globals.css (hunk 3 above)
Comment (antler-hat) [reply src/app/globals.css:59]: Pushing back - one of the goals here is design systemization, which means creating text styles that include font family, screen-responsive font size, font weight, and line height (leading). This is for repeatability, consistency, and importantly, alignment with our custom Text Styles in Figma.

Using custom classes alongside tailwind utility classes is valid and supported, and I believe we should do it. We could give them names like cd-* to indicate it's custom and not utility? eg. `cd-text-small`

Or create another text-item component(s)? Something like TextStyle?

File: src/app/globals.css (hunk 2 above)
Comment (antler-hat) [reply src/app/globals.css:117]:
```suggestion
  font-family: var(--font-body), sans-serif;
  line-height: 1.5;
```
I think we should keep these. The others are redundant because they're default browser styles anyways. I know the body rules will apply globally unless overridden but that's good and normal. There are always going to be body-level styles, if we don't set them then the browser does. May as well have this aligned with our body text style

src/app/globals.css
17cliu: i agree, we should have design systemization! i should've further expanded initially, that i would prefer to see that systemization implemented as react components -- like [these](https://github.com/civic-dashboard/civic-dashboard-web/blob/main/src/components/ui/text-items.tsx), for example. this gives better standardization than using a particular string of classnames on regular html elements, because it's too easy for someone to forget to include a classname, or to add more custom styles on top of it. of course, with components, it requires more work to make sure it works in as many contexts as possible.

PR #409, opened 2026-04-21
Strip span from agenda items summary and recs (avoid wrong text color during dark mode)

### LINKED ISSUE #408: Strip formatting html from actions data
I noticed one action today that had black text (when it should be white): https://civicdashboard.ca/actions/item/2026.EP22.9

**Card on actions page:**

Image

**Action permalink page:**

Image

Turns out this action has embedded formatting html in it:

Image

We should strip this stuff.

jeromegv:
## Description
Resolves #408

Summary of changes:
   1. src/logic/sanitize.ts: Removed span from ALLOWED_TAGS in the DOMPurify configuration but also added all related table tags as they should not be sanitized
   2. src/components/AgendaItemCard.tsx:
       * Imported sanitize.
       * Applied sanitize() to item.decisionRecommendations, item.decisionAdvice, item.agendaItemSummary, and item.agendaItemRecommendation in FullPageAgendaItemCard.
       * Applied sanitize() to item.agendaItemSummary in SearchResultAgendaItemCard.
   3. src/backend/emails/components/agendaItemCard.tsx: Imported and applied sanitize() to item.agendaItemSummary.

  These changes ensure that all HTML content sourced from the database is properly sanitized before being rendered, stripping out span tags but also only removing the color if a tag like `<p>` has a style attribute. (It was too much trouble to try to stripe every styles, we use it for padding, height, etc)

## Testing instructions

Test with this action item: 2026.EP22.9 , 2026.EC27.2 and random action items to ensure everything displays well

## Checklist

- [X ] This PR addresses all requirements described in the issue
- [ ] I have tested these changes in the preview deployment

### DIFF
--- a/src/backend/emails/components/agendaItemCard.tsx
+++ b/src/backend/emails/components/agendaItemCard.tsx
@@ -1,4 +1,5 @@
 import { AgendaItem } from '@/database/queries/agendaItems';
+import { sanitize } from '@/logic/sanitize';
 import { Heading, Link, Section, Text } from '@react-email/components';
 
 export const EmailAgendaItemCard = ({ item }: { item: AgendaItem }) => {
@@ -13,7 +14,7 @@ export const EmailAgendaItemCard = ({ item }: { item: AgendaItem }) => {
       </Heading>
       <Text
         style={{ fontSize: '16px' }}
-        dangerouslySetInnerHTML={{ __html: item.agendaItemSummary }}
+        dangerouslySetInnerHTML={{ __html: sanitize(item.agendaItemSummary) }}
       />
     </Section>
   );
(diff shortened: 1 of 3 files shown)


jeromegv:
No need to review yet, there are some fixes I still want to do, the preview link made me realize a few things.

jeromegv:
https://05181960-civic-dashboard-web.civic-dashboard.workers.dev/actions/item/2026.EP22.9
vs
https://civicdashboard.ca/actions/item/2026.EP22.9

fixes the problem

17cliu (approved):
wow i didn't know there were so many custom styles in the html 😵 

Image

question: we read these agenda items from our own db, right? could we sanitize them before storing them? that way we won't have to do as much work on the clients.

otherwise, lgtm!

jeromegv:
In theory it would make more sense to save a sanitized version in the DB, but working on the parsing is a bit of a can of worms, because its hard to know what we truly don't need vs what we need, i've noticed quite a lot of formatting depends on those html tags to make it more readable. Always the risk with HTML you don't know and don't control.
Already spent quite a few more hours than i was hoping for on this, so i'd call it good enough for now!

PR #39, opened 2025-02-22
Feat/port councillor votes page 3

mattorchard:
Opened on behalf of @sarmilan

PR #412, opened 2026-04-24
Stop using deprecated Next.js image config

17cliu:
## Description

This PR fixes the following deprecation warning when the project is started:

```
⚠ The "images.domains" configuration is deprecated. Please use "images.remotePatterns" configuration instead.
```

## Testing instructions

The preview site should build successfully and look normal.

## Checklist

### DIFF
--- a/next.config.mjs
+++ b/next.config.mjs
@@ -7,8 +7,12 @@ const nextConfig = {
         hostname: 'www.toronto.ca',
         pathname: '/wp-content/uploads/**',
       },
+      {
+        protocol: 'https',
+        hostname: 'contrib.wp.intra.prod-toronto.ca',
+        pathname: '**',
+      },
     ],
-    domains: ['contrib.wp.intra.prod-toronto.ca'],
   },
   async redirects() {
     return [


next.config.mjs
```diff
@@ -7,8 +7,12 @@ const nextConfig = {
         hostname: 'www.toronto.ca',
         pathname: '/wp-content/uploads/**',
       },
+      {
+        protocol: 'https',
+        hostname: 'contrib.wp.intra.prod-toronto.ca',
+        pathname: '**',
+      },
```
17cliu, line 14: @jeromegv do you know what images come from this location? i see that both these domains were introduced as part of councillor votes (#39), but votes don't have images (?) and the councillor images come from the other location (www.toronto.ca)

jeromegv: I don't fully know.. may be the councillor image used to come from there? But i checked our DB and it's all under toronto.ca so i don't know. Seems old and legacy for sure. IMO we could try removing it entirely and see what breaks, and likely nothing will break hehe

jeromegv (approved):

17cliu:
@jeromegv just rebuilt my local db today and noticed that we do seem to check on images in that intra host, but none of those images are valid:

image

PR #429, opened 2026-05-06
Pin versions for Github actions

17cliu:
## Description

Supply chain attacks are frequent nowadays. Our npm dependencies are pinned down via package-lock.json, but the Github actions we use are not pinned at all. We reference them by their major version number, e.g. `action/checkout@v6`, which means that each time we run the action (for example, when we create or merge a PR), it fetches the latest version of that action each time. If the action has been compromised, then the latest version could execute malicious code.

Pinning Github actions is as simple as referencing the commit SHA of the desired version, instead of the tagged version number. I've included the version number as a comment beside each SHA.

You can read more about this method here: https://dev.to/nicolasv/github-actions-why-you-should-pin-your-actions-to-a-specific-version-1513 As a side note, they also recommended enabling the "Require SHA" option in Github orgs. I think we should do that too!

## Testing instructions

CI should run successfully. 

For code review, I recommend reviewing by commit. You can reference these release tags and their commit SHAs:

- https://github.com/hoverkraft-tech/compose-action/releases/tag/v2.6.0
- https://github.com/peter-evans/create-or-update-comment/releases/tag/v5.0.0
- https://github.com/peter-evans/find-comment/releases/tag/v4.0.0
- https://github.com/actions/checkout/releases/tag/v6.0.2
- https://github.com/actions/setup-node/releases/tag/v6.4.0
- https://github.com/actions/cache/releases/tag/v5.0.5
- https://github.com/actions/github-script/releases/tag/v8

## Checklist

### DIFF
--- a/.github/workflows/check_migrations.yml
+++ b/.github/workflows/check_migrations.yml
@@ -26,7 +26,7 @@ jobs:
       # i'm not sure why `npm run db:start` doesn't work here,
       # but the remaining commands can't connect to the DB without using this action
       - name: Start database
-        uses: hoverkraft-tech/compose-action@v2
+        uses: hoverkraft-tech/compose-action@d2bee4f07e8ca410d6b196d00f90c12e7d48c33a # v2.6.0
 
       - name: Run migrations
         run: npm run db:run-migrations
--- a/.github/workflows/check_migrations.yml
+++ b/.github/workflows/check_migrations.yml
@@ -15,7 +15,7 @@ jobs:
       - name: Checkout
         uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
       - name: Setup Node
-        uses: actions/setup-node@v6
+        uses: actions/setup-node@48b55a011bda9f5d6aeb4c2d9c7362e8dae4041e # v6.4.0
         with:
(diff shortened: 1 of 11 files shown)


jeromegv (approved):

jeromegv:
Approved! Let me know once pushed, and then I can enable the feature for "Require SHA" (instructions here: https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/enabling-features-for-your-repository/managing-github-actions-settings-for-a-repository)

I'd say the only downside of that is that we lose on new security updates within the same branches.. we would need to manually update the github actions which we aren't likely to do very often. And I don't think those would show up in `npm audit`

17cliu:
> I'd say the only downside of that is that we lose on new security updates within the same branches.. we would need to manually update the github actions which we aren't likely to do very often. And I don't think those would show up in npm audit

yeah... it looks like we could configure dependabot to check for action updates: https://josh-ops.com/posts/github-dependabot-for-actions/ so we could try that! otherwise we might just have to set up a recurring task every x months to check :')

edit: opened an issue for it - #431

17cliu:
@jeromegv merged! you can enable "require SHA" now 👍

jeromegv:
@17cliu done!

PR #428, opened 2026-05-04
db: implement connection singleton to prevent exhaustion

jeromegv:
## Description
Resolves #427

Refactor `createDB` to use a global singleton pattern via `globalThis`. This ensures the Kysely instance and its underlying connection pool are  reused in DEV (but not production), preventing "too many connections" errors. Before, it was only doing it during build

## Testing instructions

Can only be tested locally by clicking on a bunch of councillor page one by one

## Checklist

- [X ] This PR addresses all requirements described in the issue
- [ X] I have commented my code, particularly in hard-to-understand areas
- [ X] I have performed a self-review of my code
- [ X] I have tested these changes in my local environment
- [ X] I have tested these changes in the preview site generated by this PR (you must finish opening the PR first)
- [ ] I have made corresponding changes to the documentation (if relevant)

PR #430, opened 2026-05-07
feat: add Cloudflare Hyperdrive support for database connections

### LINKED ISSUE #427: Opening every councillor page on the website overwhelms the database and crash the website
**Describe the bug**
Currently, on every page load, we open a new connection pool to the database. If we right-click open on every councillor page, it's enough to bring the website down (not just on the web, locally as well). Our postgre database starts refusing connections 

The createDB utility was missing a singleton pattern. In the Next.js environment, this caused a new PostgresJSDialect pool to be initialized for every server-side request. When multiple requests happen in parallel (e.g., opening multiple tabs), the cumulative connection count hits the PostgreSQL limit. This also appears to trigger a secondary TypeError in the postgres.js library's error handling logic (queryError) when it receives the connection refusal from the OS.

**To Reproduce**

Open every councillor page in a new tab, website crashes and the pages stop loading

**Expected behavior**
Open every councillor page in a new tab should open every page

jeromegv:
## Description

Integrates `@opennextjs/cloudflare` to retrieve connection strings from Hyperdrive when available. 

Limits production connection pool size to 1 to remain compatible with Cloudflare Workers' per-isolate limits.

Current state:
Every cloudflare worker in prod opens a new connection to the DB, which quickly overwhelms the DB. Because of the nature of the worker that are all independant from each other, we can't do what we do in dev environment, which is to re-use the same connection DB (which is documented here: https://github.com/civic-dashboard/civic-dashboard-web/pull/428)

Future state:

Instead we use the concept of hyperdrive. Hyperdrive is a Cloudflare tool that connects directly to our DB on a secure channel. And instead of connecting to our DB directly from our cloudflare worker, we will connect to the hyperdrive URL, and Cloudflare Hyperdrive will be in charge of querying our DB. That way we can have multiple workers that all talk to the a single Hyperdrive, which prevents overhwhelming our DB with multiple connections from multiple workers

Hyperdrive is your "Real" Pool
 In this architecture, you should think of Hyperdrive as your connection pool, and your code as a single-connection client. 

You've essentially moved the "Pooling" logic out of your application code and into Cloudflare's infrastructure. By setting max: 1 as a config for postgres, you are telling the postgres library: "Don't try to be a pool manager; just give me one stable pipe to Hyperdrive, and let Hyperdrive handle the rest."

Resolves #427 

## Testing instructions

Go to the councillor page on the preview link, open a bunch of councillors quickly, ensure the website does not crash. It can still get overwhelmed but it seems to require more councillors page before it goes down, as hyperdrive seems like a better pattern anyway.

## Checklist

- [ X] This PR addresses all requirements described in the issue
- [X ] I have commented my code, particularly in hard-to-understand areas
- [X ] I have tested these changes in my local environment
- [X ] I have tested these changes in the preview site generated by this PR (you must finish opening the PR first)
- [X ] I have made corresponding changes to the documentation (if relevant)


jeromegv:
Haven't tested yet, not ready for review.

src/database/psql.ts
```diff
@@ -1,9 +1,24 @@
...
+    // getCloudflareContext may fail in non-Cloudflare environments (e.g. during build or local scripts)
+  }
+
   if (!connectionString)
-    throw new Error(`No value for "DATABASE_URL" provided`);
+    throw new Error('No value for "DATABASE_URL" or "HYPERDRIVE" provided');
 
-  return postgres(connectionString);
+  return postgres(connectionString, {
+    max: process.env.NODE_ENV === 'production' ? 1 : undefined,
```
garobrik, line 22: why do we need to limit the size of the connection pool in the Worker now that we're using Hyperdrive?

src/database/kyselyDb.ts
```diff
@@ -3,16 +3,24 @@ import { PostgresJSDialect } from 'kysely-postgres-js';
 import { createPostgres } from '@/database/psql';
 import { DB } from '@/database/allDbTypes';
 import { PHASE_PRODUCTION_BUILD } from 'next/dist/shared/lib/constants';
+import { cache } from 'react';
 
 const globalForDb = globalThis as unknown as {
   __kysely_db__: Kysely<DB> | undefined;
 };
```
garobrik, line 10: could we just rely on `cache` for all of our singleton needs now? do we still need to do it this way for local dev/static generation?

garobrik (approved):
no blocking comments, but i left a couple questions i'm curious about!

garobrik:
ok wait i approved this PR, but now i went back to test it and can still get an error by opening lots of tabs, and then i looked into it some more:
-  on civicdashboard.ca, the error i get when i open many tabs is "Worker exceeded resource limits", but on the preview branch i can also get "uncaught exception". but there's no way to see what the uncaught exception is in the Cloudflare dashboard for preview deployments :/ 
- i also noticed on the Sevalla dashboard that the DB memory usage spiked above the limit when i opened a lot of tabs, so maybe we've traded one problem for another? (Sevalla instance only has like 250mb of memory which is not a lot, the next tier is $34usd/mo for 2GB of memory)
- re: "worker exceeded resource limits", we only get 10ms of CPU time per request on the free tier, whereas we get up to 5m when we get on the $5/mo plan, so i'd be curious to see whether that would resolve it too
- i wonder if there's anything we can do also to trace where all the CPU usage is, because if the issue is number of connections, that shouldn't cause CPU spikes :thinking: (unless it's just retrying over and over when the DB is unavailable and the retries are what is eating all the CPU)

17cliu:
i'm not familiar with hyperdrive, but also curious to see responses to garo's comments 👀 

i used the preview to open a bunch of councillor pages. after around 20, i hit a `Worker threw exception` error. in case you want to look it up in the logs: 
- bbc0b244-civic-dashboard-web.civic-dashboard.workers.dev
- `Error 1101 • Ray ID: 9fa2c0d49b3aac34 • 2026-05-11 17:02:48 UTC`

src/database/psql.ts
jeromegv: ```You've essentially moved the "Pooling" logic out of your application code and into Cloudflare's infrastructure. By setting max: 1 as a config for postgres, you are telling the postgres library: "Don't try to be a pool manager; just give me one stable pipe to Hyperdrive, and let Hyperdrive handle the rest."```
I'm not 100% sure if this is the right approach, but it was suggested to me by AI and it seems to make sense. And i'm wondering if that's why we don't hit CPU overload on the preview branch (and hit another type of exception instead)
But it's just a theory because of course, we don't have the logs.

jeromegv:
Unfortunately t[here are no logs when we work in Preview](https://developers.cloudflare.com/workers/configuration/previews/) (very annoying) so it's hard to know for sure. So it's kind of hard to improve without deploying to prod and doing real tests there. Even then, sometimes logs aren't super helpful.

And yes, it does seem to still crash, but it seems that it requires more tabs before it crashes.. so we somewhat improve things. At least that was my testing. May be i'm imagining things?

```
i also noticed on the Sevalla dashboard that the DB memory usage spiked above the limit when i opened a lot of tabs, so maybe we've traded one problem for another? (Sevalla instance only has like 250mb of memory which is not a lot, the next tier is $34usd/mo for 2GB of memory)
```

Didn't think of looking at things there, it's definitely interesting.

I've also noticed Hyperdrive has a limit of 20 connections to Sevalla, that might be too many, opened a support ticket with Sevalla to ask what they recommend as limit on our free plan
Screenshot 2026-05-11 at 5 59 05 PM

Let's wait to see what Sevalla says, and then we can decide if we are comfortable with at least deploying in prod and have better access to logs. Because may be we can catch an exception that tells us that there are too many connections and we can fail in a better way (which is an improvement of prod failing with too many CPU exception)

Overall, from an architecture perspective, it still seems to be a better architecture because hyperdrive will auto-cache queries and it's a more standard pattern to talk to a database from a worker perspective, so thats what i was trying to improve, even if it didn't end up being perfect. But if it's too much fucking around, i might just drop it entirely.

jeromegv:
So this was the answer from Sevalla support
```👋 For a 250 MB Postgres plan, keep your connection pool capped at 10 to 15 connections max. That plan has very little RAM, and each Postgres connection eats roughly 5 to 10 MB of working memory once active, so going higher risks running the database out of memory before you ever hit the official connection limit. Also set an idle timeout of 30 to 60 seconds on the pool so unused connections release quickly, and leave some headroom for backups, health checks, and the Sevalla Database Studio.```

So I changed Cloudflare Hyperdrive to a max of 10 connections to be safe
Screenshot 2026-05-12 at 5 10 22 PM

And then, I opened all the councillors twice (!)

NO CRASH!

Sevalla postgres caps the memory usage just under the limit, but it never breaches it
Screenshot 2026-05-12 at 5 11 24 PM

All that happens is that the list of agenda items take longer to load on the councillor page. We can see in those metrics that essentially Hyperdrive just queues those queries and wait longer to send them to the database, but beside a slower loading time, it does not influence anything else. The website keeps serving.

Screenshot 2026-05-12 at 5 12 48 PM
Screenshot 2026-05-12 at 5 12 42 PM

jeromegv:
Decided to merge since the preview testing was quite successful and i figured that the real testing would be in prod anyway (and would rather do it when i had time to deal with the fallback if things went badly)

Did have to do an emergency PR but that was not due to performance
https://github.com/civic-dashboard/civic-dashboard-web/pull/438

garobrik:
@jeromegv that's awesome, thanks for figuring this out!!

PR #441, opened 2026-05-19
Fix scroll behaviour on decision body filter

### LINKED ISSUE #355: Decision body search should automatically show 1st item
### Description
When doing a search on the actions page within the decision body dropdown, doesn't show the first results of that search, it shows some items down the list. It should automatically show the first item in the dropdown list

Image

Example:
Go to the actions page
Go to decision body dropdown
Type transit

Some other committee is shown, instead of one of the TTC decision body (we have to manually scroll back up to select them)

joykare:
## Description

Resolves #355

**Bug:** Scroll remained in an arbitrary position after dropdown is reordered on sort
**Expected:** Scroll to move to the top if dropdown supports reordering 

**BEFORE:**

https://github.com/user-attachments/assets/228ca517-230b-4d17-a7b5-45559f06a295

**AFTER:**

https://github.com/user-attachments/assets/8107ae95-221f-46a7-8da8-76a5befbda83

## Testing instructions

1. Navigate to Action page
2. On the decision body filter Search "transit"
3. Scroll should remain at the top
4. On the Dropdown next to the Search bar make sure behaviour is not changed when you select "Newest" | "Oldest" | "Most relevant" 

## Checklist

### DIFF
--- a/src/components/search.tsx
+++ b/src/components/search.tsx
@@ -106,6 +106,7 @@ export function DecisionBodyFilter({
       value={decisionBodyIds}
       onSelect={onSelect}
       placeholder="Select decision body..."
+      scrollToTopOnSearch
     />
   );
 }
--- a/src/components/search.tsx
+++ b/src/components/search.tsx
@@ -63,7 +63,7 @@ export function SortDropdown() {
       placeholder="Sort by..."
       multiple={false}
       searchable={false}
-      reorderSelected={false}
+      reorderListItems={false}
     />
   );
 }
@@ -106,7 +106,6 @@ export function DecisionBodyFilter({
       value={decisionBodyIds}
       onSelect={onSelect}
       placeholder="Select decision body..."
-      scrollToTopOnSearch
     />
   );
 }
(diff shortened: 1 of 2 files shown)


jeromegv:
Functionally in the video it seems to fix the problem, and on the preview link it seems to work as well, however for the code review i'll leave it to @17cliu or Karishma!

17cliu (approved):
works well! thanks for finding and fixing this 🎉 

one thought: do we need `resetScrollOnSearch` to be an option? in what situation would we not want the scroll to reset when the search query changes? 🤔 i think it's reasonable for the scroll to reset when there's a new search term... my opinion is that maybe we can just make it the default behaviour, but i don't feel strongly about it. feel free to merge either way!

joykare:
Thanks @17cliu - so the reason I didn't make it default is because we use the combobox for both the Select and Search. For the Select it looks like we typically dont reorder the options but on search we do. Hence the optional reset scroll on value change.

jeromegv:
Let's get it merged! @joykare

PR #448, opened 2026-06-02
fix/382 - partial matching of action item code

### LINKED ISSUE #382: Searching for a partial action item code has 0 results
E.g. on the actions page, I cannot search "sc28" to find action item `2026.SC28.1`. "sc28" has zero results. I have to search the full action item code to find the item.

imadsyed333:
## Description

Resolves #382

Implements partial matching of text query with the reference field of action items.

Before, inputting substrings of action item codes would yield no results. Now, action items with codes that partially match the user's query are returned.

Before:
image

After:
image

## Testing instructions

Go to the Actions page and search for parts of action item codes (e.g. "GG", "CA36"). You should see a list of action items with codes that partially match the search query.

Also, try searching for key terms and selecting tags to ensure that the new logic doesn't interfere with existing search functionality.

## Checklist

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -8165,7 +8165,6 @@
       "integrity": "sha512-tVQRucExLQ02Boi4vdPp49svNGcfL2GhdTCT9aldhXgCJVAI21EtRfBettiuLUwce/7r6bFdgs6JFkcdTiFttA==",
       "dev": true,
       "license": "MIT",
-      "peer": true,
       "dependencies": {
         "@ampproject/remapping": "^2.2.0",
         "@babel/code-frame": "^7.24.2",
@@ -8685,7 +8684,6 @@
         }
       ],
       "license": "MIT",
-      "peer": true,
       "engines": {
         "node": ">=18"
       },
@@ -8708,7 +8706,6 @@
         }
       ],
       "license": "MIT",
-      "peer": true,
       "engines": {
         "node": ">=18"
       }
@@ -10013,7 +10010,6 @@
       "integrity": "sha512-rONPWMC7PeExE077uLE4oqWrZ1IvAfz3oH9LibVAcVCopJiA9R62uavnbEzdkVmJYI6M6Zgkbeb07+tWjlq2XA==",
       "dev": true,
       "license": "MIT",
-      "peer": true,
       "engines": {
         "node": "^14.21.3 || >=16"
       },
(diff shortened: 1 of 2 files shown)


jeromegv:
@imadsyed333 thank you for your PR!

1. The package lock update does not seem related to this PR, can you remove it?

2. So my AI noticed that:

```
  File: src/database/queries/agendaItems.ts
  L391: [HIGH] Search and tag filters are broken when textQuery is empty.

  If postgresQuery is truthy (e.g., when filtering by tags) but textQuery is an empty string, the condition eb('reference', 'ilike', '%%') will always return true.
  Because this is joined with an OR, the entire filter is bypassed, and all agenda items will be returned regardless of the selected tags or full-text search
  requirements.

  Suggested change:

    1 -        query = query.where((eb) =>
    2 -          eb.or([
    3 -            sql<boolean>`"textSearchVector" @@ query.query`,
    4 -            eb('reference', 'ilike', `%${textQuery}%`),
    5 -          ]),
    6 -        );
    7 +        query = query.where((eb) => {
    8 +          const orConditions = [sql<boolean>`"textSearchVector" @@ query.query`]
    9 +          if (textQuery.trim() !== '') {
   10 +            orConditions.push(eb('reference', 'ilike', `%${textQuery}%`))
   11 +          }
   12 +          return eb.or(orConditions)
   13 +        });

```
I'm not saying to follow the code suggestion, but there is indeed a bug
if you go there:
https://fdba1074-civic-dashboard-web.civic-dashboard.workers.dev/actions
Keep an empty text query
Then filter with any tag, it doesn't filter anymore, it always returns 100% of the results
I don't see this behavior on 
https://civicdashboard.ca/actions

So that seems like a regression/bug

imadsyed333:
Thanks for the feedback!

1. the package lock change showed up as soon as i ran "npm install". Should be easy to remove, and I'll use "npm ci" moving forward.

2. Yeah, I see the actions being re-ordered instead of being filtered. Will patch this as soon as possible!

imadsyed333:
@jeromegv 
1. Package-lock changes have been removed.

2. I first implemented a variation of the code changes your AI suggested, and it solved the empty query + tag issue, however I soon discovered another issue:

If you type in part of an action item code (eg. PB45) and select a tag (eg, housing), you will get results matching either the text query OR the tags (which makes sense given the code), but this doesn't align with how tags + text query work in production. 

So now i have added new variables: combinedQuery (this replaces postgresQuery), tagsQuery (tags only query string), and textOnlyQuery (text only query string). 

if tagsQuery is not empty, tags exist and the query builder object adds a query to match tags through textVectorSearch

if textOnlyQuery is not empty, text has been inputted and so the query builder object adds, textVectorSearch as well as partial match on action item code.

If both tags and text exist, then the query builder object will have queries from the above cases, intersecting results from both queries.

Let me know what you think!

jeromegv (approved):

jeromegv:
I think this looks great, you can merge!

imadsyed333:
Thanks, shipping it 🚀!

PR #449, opened 2026-06-04
Create new text component to handle text styles with semantic control

antler-hat:
## Description

This evolves and consolidates the current heading heading components into a single Text component that can be set to different style presets AND optionally given an alternative semantic role for a11y best practices

It also implements the specific text styles of our Figma design system

This slightly tweaks the text styles from where they are right now, including adding some screen responsive size changes. It is mostly imperceptible but should reflect [what we have in Figma](https://www.figma.com/design/cGSzuLZ94FKDvbxmucEqVs/Civic-Dashboard?node-id=1605-521&t=XDqRxQgY2yAw4mCy-11)

Preview the text items in action here: https://1ab2e046-civic-dashboard-web.civic-dashboard.workers.dev/labs/text-items

## Checklist

### DIFF
--- a/src/app/labs/text-items/page.tsx
+++ b/src/app/labs/text-items/page.tsx
@@ -47,7 +47,7 @@ function VariantCard({ preset }: { preset: TextPreset }) {
 
   return (
     <Card>
-      <CardHeader className="border-b border-gray-light bg-gray-lightest flex-col md:grid md:grid-cols-4 gap-1">
+      <CardHeader className="border-b border-gray-light flex-col md:grid md:grid-cols-4 gap-1">
         <Text preset="Heading3" className="mb-0 text-lg">
           {preset}
         </Text>
--- a/src/app/labs/text-items/page.tsx
+++ b/src/app/labs/text-items/page.tsx
@@ -1,0 +1,127 @@
+import { Metadata } from 'next';
+import { Card, CardContent, CardHeader, CardTitle } from '@/components/ui/card';
+import { Page } from '@/components/ui/page';
+import { Section } from '@/components/ui/section';
+import {
+  BulletedList,
+  TEXT_PRESETS,
+  Text,
+  textPresetElements,
+  type TextPreset,
+} from '@/components/ui/text-items';
+
+export const metadata: Metadata = {
+  title: 'Text Items Lab',
+  description:
+    'Reference page showing every shared Text and BulletedList variant.',
+};
+
+const presetCopy: Record<
+  TextPreset,
+  { sample: string; usage: string; wrapperClassName?: string }
+> = {
+  Heading1: {
+    sample: 'Democracy should feel legible before it feels impressive.',
+    usage: 'Primary page or hero heading.',
+  },
+  Heading2: {
+    sample:
+      'Use section headings to break dense civic content into readable blocks.',
+    usage: 'Major section title.',
+  },
+  Heading3: {
+    sample: 'Smaller callouts work well for supporting subsections.',
+    usage: 'Subsection heading inside a larger narrative.',
+  },
+  Body: {
+    sample:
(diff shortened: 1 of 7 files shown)


antler-hat:
@17cliu what do you think of this approach? I'm trying for a clean way to handle consistent  text styles while being flexible with semantics

17cliu (commented):
hey! sorry for the late reply -- i've been rolling this around in my mind a little 😅 

this is an interesting approach! i like how it rolls everything into one element -- i think that would be helpful for code editor autocomplete. it would make the available presets easy to find. however, i also worry if it would make maintenance difficult -- it makes the one `Text` component "load-bearing". we would have to be careful to watch for additions of complicated behaviour in the future (e.g. if we had to add code like "if h3 and this-context, then do this, else do that"), and break that out into other components.

i'd also like a second opinion from another dev too -- @joykare, what do you think?

src/components/ui/text-items.tsx
```diff
@@ -1,31 +1,74 @@
...
+  children: ReactNode;
+  className?: string;
+} & Omit<HTMLAttributes<HTMLElement>, 'children' | 'className'>;
+
+export const textPresetElements: Record<TextPreset, TextElement> = {
+  Heading1: 'h1',
+  Heading2: 'h2',
+  Heading3: 'h3',
+  Body: 'p',
+  Small: 'p',
```
17cliu: recommend using a real [`small`](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/small) tag for this, so that if there are any default semantic behaviour for it, it'll get automatically applied.
```suggestion
  Small: 'small',
```

antler-hat:
@17cliu I'm easy. This approach is like MUI i believe. A counter example would be Chakra UI where each HTML typography element has its own component.

src/app/labs/text-items/page.tsx
```diff
@@ -0,0 +1,127 @@
...
+  },
+  Heading3: {
+    sample: 'Smaller callouts work well for supporting subsections.',
+    usage: 'Subsection heading inside a larger narrative.',
+  },
+  Body: {
+    sample:
+      'Body text is the default reading style for explanatory copy, descriptions, and most page content where clarity matters more than emphasis.',
+    usage: 'Default long-form paragraph style.',
+    wrapperClassName: 'max-w-2xl',
```
jtrang: Is `wrapperClassName` here (and in line 46) supposed to be used in the examples in the `TextItemsLabPage` component?

src/components/ui/text-items.tsx
```diff
@@ -1,31 +1,74 @@
...
+  | 'span';
 
-/**
- * Renders a `<h3>` with correct styling.
- */
-export function Heading3({ children }: { children: React.ReactNode }) {
-  return <h3 className="text-xl font-bold mb-6">{children}</h3>;
+type TextProps = {
+  preset: TextPreset;
+  as?: TextElement /* Optionally override the default HTML element for a11y compliance */;
```
jtrang: Personally, I find the `as` prop name a little obscure. I think a more descriptive name would be helpful here.

jtrang (commented):
I'm not familiar enough with the codebase yet to have a strong opinion either way, but it makes sense to me to consolidate the presets in the `Text` component if the design system is set on them.

Minor note: on the lab page, the table headings overlap on smaller screen sizes (under 440px). Screenshot attached.

Image

src/app/labs/text-items/page.tsx
antler-hat: fixed 👍

src/components/ui/text-items.tsx
antler-hat: Changed to `tag` which would be more accurate.
Open to other ideas!

jtrang (approved):
Looks good to me :)

PR #451, opened 2026-06-08
Enhance `EmailAgendaItemCard` with improved styling, date formatting, and status display.

### LINKED ISSUE #46: Implement agenda item card for email template
Unfortunately you can't just stick normal HTML in an e-mail template, so we'll need to implement the agenda item card twice, once for individual pages and once for emails. See `sendSearchResultsEmail.tsx` for the current implementation, and [react-email](https://github.com/resend/react-email) for documentation on our templating library. There is [very nascent documentation](https://github.com/resend/react-email) on our codebase's (lack of) email DX.

Onuiteshi:
Description
Resolves #46

The EmailAgendaItemCard component previously only showed the item reference, title, and summary. This PR enhances it to match the information hierarchy of the website card.

Before: Reference + title (as a link) + plain summary text only.

After: Meeting date, decision body, title (as a styled link), status badge (when present), decision recommendations (when present), and summary  all inside a card with a left accent border, subtle container border, and consistent typography.

**Testing instructions**
**Local:**
1. Run `npm run docker:start && npm run dev`
2. Go to `/actions`, search "housing"
3. Click Get Email Alerts, enter any email
4. Check maildev at http://localhost:1080

**Preview deployment:**
1. Open the preview URL and go to `/actions`
2. Search any topic, click Get Email Alerts, enter your real email
3. Check your inbox for the subscription confirmation email

Checklist

### DIFF
--- a/src/backend/emails/components/agendaItemCard.tsx
+++ b/src/backend/emails/components/agendaItemCard.tsx
@@ -29,9 +29,7 @@ export const EmailAgendaItemCard = ({ item }: { item: AgendaItem }) => {
           {item.reference}: {item.agendaItemTitle}
         </Link>
       </Heading>
-      {formattedStatus && (
-        <Text style={statusBadge}>{formattedStatus}</Text>
-      )}
+      {formattedStatus && <Text style={statusBadge}>{formattedStatus}</Text>}
       <Hr style={divider} />
       {item.decisionRecommendations && (
         <>
--- a/src/backend/emails/components/agendaItemCard.tsx
+++ b/src/backend/emails/components/agendaItemCard.tsx
@@ -1,7 +1,14 @@
 import { AgendaItem } from '@/database/queries/agendaItems';
-import { sanitize, stripHtmlAndGetFirstParagraph } from '@/logic/sanitize';
+import { stripHtmlAndGetFirstParagraph } from '@/logic/sanitize';
 import { formatAgendaItemStatus } from '@/logic/strings';
-import { Button, Heading, Hr, Link, Section, Text } from '@react-email/components';
+import {
+  Button,
+  Heading,
+  Hr,
+  Link,
+  Section,
+  Text,
+} from '@react-email/components';
 
 const SUMMARY_MAX_CHARS = 200;
 
--- a/src/backend/emails/components/agendaItemCard.tsx
+++ b/src/backend/emails/components/agendaItemCard.tsx
@@ -1,21 +1,93 @@
 import { AgendaItem } from '@/database/queries/agendaItems';
 import { sanitize } from '@/logic/sanitize';
-import { Heading, Link, Section, Text } from '@react-email/components';
+import { formatAgendaItemStatus } from '@/logic/strings';
+import { Heading, Hr, Link, Section, Text } from '@react-email/components';
+
+const dateFormatter = new Intl.DateTimeFormat('en-US', {
+  month: 'short',
+  year: 'numeric',
+  day: 'numeric',
+  timeZone: 'America/Toronto',
+});
 
 export const EmailAgendaItemCard = ({ item }: { item: AgendaItem }) => {
+  const formattedDate = dateFormatter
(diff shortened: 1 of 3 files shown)


src/backend/emails/templates/subscriptionUpdate.tsx
```diff
@@ -1,11 +1,13 @@
...
 import { EmailWrapper } from '@/backend/emails/components/emailWrapper';
 import { EmailSubscriptionCard } from '@/backend/emails/components/subscriptionCard';
 import { EmailAgendaItemCard } from '@/backend/emails/components/agendaItemCard';
-import { SubscribableSearchFilters } from '@/logic/search';
 import { UnsubscribeLink } from '@/backend/emails/components/unsubscribeLink';
+import { SubscribableSearchFilters } from '@/logic/search';
+import { Heading, Hr, Link, Section, Text } from '@react-email/components';
+import { Fragment } from 'react';
+
+const MAX_ITEMS = 2;
```
17cliu: just wanted to confirm -- this is on purpose? we want to limit the update emails to just two cards and force readers to click through to the website?

17cliu (approved):
wow, this looks really good! 🌟

|before| after|
|-|-|
|Image|Image

 apologies for the late review on this. just two questions; otherwise, lgtm!

src/backend/emails/components/agendaItemCard.tsx
```diff
@@ -1,21 +1,90 @@
...
+      <Heading as="h2" style={heading}>
+        <Link href={itemUrl} style={titleLink}>
           {item.reference}: {item.agendaItemTitle}
         </Link>
       </Heading>
-      <Text
-        style={{ fontSize: '16px' }}
-        dangerouslySetInnerHTML={{ __html: sanitize(item.agendaItemSummary) }}
-      />
+      {formattedStatus && <Text style={statusBadge}>{formattedStatus}</Text>}
```
17cliu: hmm, i think it'd be nice to prefix this with `Status: `, as we do on the website. otherwise, it could be confusing. for example with the "no action required" status, readers might think this is saying that there is no action required of _them_, rather than that being just what the council concluded.

Image

jeromegv: Great point!

src/backend/emails/templates/subscriptionUpdate.tsx
jeromegv: I don't think we have a limit right now? I think most people would not expect a limit (or would expect a much higher limit)

src/backend/emails/components/agendaItemCard.tsx
Onuiteshi: Yea, I have added the "Status:" prefix to match the website card.

src/backend/emails/templates/subscriptionUpdate.tsx
Onuiteshi: Yes this was intentional. Previously there was no limit so a popular tag like "housing" could dump hundreds of items into a single email which isn't a great experience. I bumped it to 5 as a reasonable balance between giving enough context and not overwhelming the inbox.

Also worth noting, the items previously had no guaranteed order so older ones could show up at the top. They now show the most recent first so subscribers always see what's newest. The "And X more results" link at the bottom takes them to the site to see everything.
Was there a specific reason the previous behaviour had no limit? If the team thinks 5 is still too low I can take it higher

@17cliu @jeromegv

jeromegv: > Yes this was intentional. Previously there was no limit so a popular tag like "housing" could dump hundreds of items into a single email which isn't a great experience.  

This might be because of the testing in the dev environnment. Once a agenda item has been sent by email to subscribers, we switch it the  value of `notificationSent` to true, and it doesn't get sent again. So normally, it's only the "new" housing element of a single day that is sent, so its not hundreds, but more like just a few.

I'll have few more comments, BRB.

PR #452, opened 2026-06-09
fix-email-search

### LINKED ISSUE #382: Searching for a partial action item code has 0 results

imadsyed333:
## Description

Resolves #382

Implements partial matching for codes of action items that users are subscribed to. Brings subscription agenda item search to parity with the main agenda item search functionality.

## Testing instructions

Go to the "Actions" page, and submit any combination of search query + selected tags. I'd focus on partial action item code searches + tags.

Then, subscribe for email notifications. The list of action items in your email should match the search results on the "Actions" page.

## Checklist

### DIFF
--- a/src/database/queries/agendaItems.ts
+++ b/src/database/queries/agendaItems.ts
@@ -514,6 +514,20 @@ export const getSubscribersToNotify = async (db: Kysely<DB>) => {
                 '@@',
                 eb.ref('Subscriptions.tsQuery'),
               ),
+              // NEW: Also match partial reference like new search does
+              eb.and([
+                eb('Subscriptions.textQuery', '!=', ''),
+                sql<boolean>`"newConsiderations"."reference" ILIKE '%' || "Subscriptions"."textQuery" || '%'`,
+                // Ensure tags also match if present (to avoid false positives)
+                eb.or([
+                  sql<boolean>`array_length("Subscriptions"."tags", 1) IS NULL`,
+                  eb(
+                    'newConsiderations.textSearchVector',
+                    '@@',
+                    eb.ref('Subscriptions.tsQuery'),
+                  ),
+                ]),
+              ]),
             ]),
           )
           .where('Subscribers.email', '!=', '')


jeromegv:
Might be an issue to rely on tsQuery

Yes, PR 448 addressed the same functional requirement (partial reference matching) but it did not have this issue because it was implemented in the searchAgendaItems
  function, where queries are generated on the fly. 

  In PR 448, the author correctly separated the tags from the text:
   1. They created a tagsQuery (tags only).
   2. They created a textOnlyQuery (text only).
   3. They applied them as independent conditions in the SQL WHERE clause.

  PR 452 tries to replicate this logic in getSubscribersToNotify, but it fails because it relies on the Subscriptions.tsQuery column. In our database, tsQuery is a
  pre-computed combination of both tags and text (see src/database/queries/subscriptions.ts).

  Why PR 452 is broken (and PR 448 wasn't):

  If a subscriber follows the tag "Housing" and the search term "EX1.1":
   * PR 448 (Search): It checks if the item has the "Housing" tag AND (matches "EX1.1" in text OR "EX1.1" in reference). This works perfectly.
   * PR 452 (Notifications): It checks if the item matches the reference "EX1.1" AND (matches the combined query housing & ex1.1 in the text vector). If "EX1.1" only
     exists in the reference field and not the summary/title, the tsSearchVector will not match the combined query, and the user gets no notification.

  To fix this properly in PR 452, we would need to either:
   1. Store tagsTsQuery and textTsQuery as separate columns in the Subscriptions table.
   2. Change the notification query to parse the tags array into a fresh tsquery on the fly, similar to how PR 448 does it for search.
   
---
I don't think a database change is in order... not sure how much work is it for #2 
Be advised as well that we will be changing tags search soon, so i definitely dont want you to spend too much time on that if this become a can of worms..

imadsyed333:
Yeah, my AI agent was giving me a similar warning as well, but when I tried a similar example (search "IE30.1" + tag "traffic") both locally and in the preview build, the email I received contained the same action items as the search results. So I'm not sure what it's trying to say here.

Interestingly, when I try this same example in production (or remove the tags to replicate the original bug), the results in the email match the results in the "Actions" page.

jeromegv:
Hi @imadsyed333 , i didn't forget you, i just need some time to focus on that, it wasn't my code originally so i need some time to ramp up on that and truly understand. Sorry for the delay!

PR #454, opened 2026-06-23
Improve agenda item search to support highlighting and sanitize output. Update components and types accordingly.

### LINKED ISSUE #60: Improve results highlighting for text search
Currently we have a pretty basic solution for highlighting users' searches in the results, based on a client-side library called [mark.js](https://markjs.io/). It doesn't use the same search logic as the backend, instead searching for exact substring matches. So for example if you search for "dogs", the backend will find results with the word "dog", but they won't be highlighted. And conversely, if you search for "con", we will highlight words like "uncontroversial" in the results even though they aren't actually what the backend matched on. 

Postgres provides a function called [`ts_headline`](https://www.postgresql.org/docs/current/textsearch-controls.html#TEXTSEARCH-HEADLINE) which can provide abbreviated and highlighted results from a text document, similar to what you might find in a search engine results page. I've never used it before, so I'm not sure how much work it'll be to integrate it, and whether it will have trouble with the fact that the text we're returning is HTML formatted.

Onuiteshi:
## Description

Resolves #60

Search results previously used mark.js on the client to highlight matched terms with exact substring matching. This conflicted with the backend's Postgres full-text search, which uses English stemming so searching "housing" would find items containing "housed" or "houses" but mark.js would fail to highlight them, or would highlight unrelated substrings.

Before: Full HTML summary rendered in each card, client-side mark.js highlights exact substrings only. Stemmed matches (e.g. "dog" matching "dogs") were never highlighted.

After: Postgres ts_headline generates a short relevant excerpt server-side, with matched terms wrapped in <mark> tags that align exactly with what the full-text search engine matched. Tag-only searches (no text query) continue to show the full summary as before.

## Testing instructions

Local:
Run npm run docker:start && npm run dev
Go to /actions and type a search term like "park" or "zoning"
Confirm that stemmed variants are highlighted e.g. searching "parks" should highlight "parking" and "park" in the snippet
Click a tag without typing anything cards should show the full summary with no highlights
Preview deployment:

Open the preview URL and go to /actions
Search any topic and verify highlighted excerpts appear in the result cards
Confirm tag-only filtering still shows full summaries

** Confirm that stemmed variants are highlighted**

| before | after |
|-|-|
| image | image |
| Full HTML summary rendered. Client-side mark.js highlights exact substrings only — stemmed matches are never highlighted. | Postgres `ts_headline` returns a short relevant excerpt with matched terms highlighted in yellow, aligned with what the search engine actually matched. |

## Checklist

### DIFF
--- a/src/app/api/agenda-item/search/route.ts
+++ b/src/app/api/agenda-item/search/route.ts
@@ -1,7 +1,10 @@
 import { parseDateParam, parseNumberParam } from '@/api/utils';
 import { isTag } from '@/constants/tags';
 import { createDB } from '@/database/kyselyDb';
-import { AgendaItem, searchAgendaItems } from '@/database/queries/agendaItems';
+import {
+  AgendaItemSearchResult,
+  searchAgendaItems,
+} from '@/database/queries/agendaItems';
 import {
   SortByOption,
   sortByOptions,
@@ -14,7 +17,7 @@ export type AgendaItemSearchResponse = {
   page: number;
   pageSize: number;
   totalCount: number;
-  results: AgendaItem[];
+  results: AgendaItemSearchResult[];
 };
 
 export async function GET(request: NextRequest) {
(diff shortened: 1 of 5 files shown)


jeromegv:
Wow! Amazing. It's a good approach. Approved!

jeromegv (approved):

jeromegv:
@Onuiteshi you can merge

PR #471, opened 2026-07-31
fix-426/update-committee: GitHub Actions for decision bodies

### LINKED ISSUE #426: Decision Bodies (committees) should be updated by data pipeline, not be static in code
### Decision Bodies (Committees)

TMMIS provides the list of decision bodies, which we use to filter agenda items on the Actions page and also to list the committes on the councillor page

* **Data Schema:** [`src/api/decisionBody.ts`](../src/api/decisionBody.ts)
* **Storage:** Saved statically in [`src/constants/decisionBodies.ts`](../src/constants/decisionBodies.ts) (not currently stored in the database).
* **Updates:** Updates are manual. Run the [`src/scripts/fetchDecisionBodies.ts`](../src/scripts/fetchDecisionBodies.ts) script to fetch a new JSON list for each council term and manually update the constant file.

## Future state:
This should be automatically be updated through our data pipeline (Github Actions) without having to manually update code whenever a committee member change.

imadsyed333:
## Description

Resolves #426

Before, decision bodies were stored in the static `decisionBodies.ts` file that had to be updated manually via the console output of `fetchDecisionBodies.ts`. 

Now, a weekly GitHub action will run `fetchDecisionBodies.ts`, which will automatically generate `decisionBodies.ts`. If there are changes, a pull request is created to merge into `main`.

I would like feedback regarding the update process (e.g, Should a PR be created on decision bodies change?).

## Checklist


mintybasil:
@imadsyed333 New here, still getting a handle on things 👋  

Wondering if it has been discussed to move the decision bodies into the DB? Maybe a future PR, this just made me realize they are not actually constant 😅

imadsyed333:
> @imadsyed333 New here, still getting a handle on things 👋
> 
> Wondering if it has been discussed to move the decision bodies into the DB? Maybe a future PR, this just made me realize they are not actually constant 😅

Hi! Its been discussed (sort of ), and ideally, it would be much cleaner to have it in a DB. The concern is mainly the sheer amount of decision bodies, and how often they are needed. So there could be issues with storage + rate limits (caching could help with this, though) if we do use a DB.

I see this as more of a temporary fix until a full solution is drawn out.

jeromegv:
I don't think i'm concerned about storage really fyi!

We would just need to evaluate if this is a big hit to hit the DB instead of our code to get that data in terms of response time, but that fear might be unwarranted, we just never evaluated it.

imadsyed333:
> I don't think i'm concerned about storage really fyi!
> 
> We would just need to evaluate if this is a big hit to hit the DB instead of our code to get that data in terms of response time, but that fear might be unwarranted, we just never evaluated it.

Ah, my apologies. In that case, we could give the decision bodies in DB a shot. I've just pushed a slew of changes that migrate decision bodies into the kysely db and provide query functions wherever decision bodies were needed. These also utilize caching to limit the number of times we fetch from the DB.

However, since there is a new decision bodies table for db, I believe the db used in the preview deployment will have to be migrated.

mintybasil:
@imadsyed333 I think we can close this?

PR #474, opened 2026-08-03
feat-146: Subscription email preview

### LINKED ISSUE #146: feat: better communicate to users what they're signing up for when they subscribe to email alerts
### Description

Currently, when a user subscribes to email alerts at [civicdashboard.ca/actions](https://civicdashboard.ca/actions), the only explanation they get of what they'll receive upon signing up is text saying "You will be sent emails for new agenda items matching your current tags & search filters, and can unsubscribe at any time." _Once_ they sign up, they'll receive an email saying "this is what the emails you'll receive will look like".

This can be confusing, and if there are no current items matching the users' search it's also hard for them to imagine what they might be receiving. To improve on this, we could:
- Make the pop-up full screen, and show an email preview inline in the page
- Automatically display results from the past when there are no current results for the search

imadsyed333:
## Description

Resolves #146 

Before, subscribing to agenda item search results would not provide users with an idea of what the email would look like. Now, users are shown a preview of the email current results (or if none, then past results) matching their search criteria. 

If the search criteria is too restrictive, users will be notified with a message under the email input.

image

image

## Testing instructions

1. Enter a search query, a bunch of tags, and/or decision bodies
2. Press the "Get Email Alerts" button. You should see an almost fullscreen modal  appear with a place to input email and subscribe, as well as a preview of the email you will receive. If your search filters are too restrictive, you should also see a message notifying you of that.
3. Pressing the "Subscribe" button should send you the email you were shown.

## Checklist


src/backend/emails/sampleAgendaItems.ts
garobrik, line 1: nit: i think we can remove this file from the PR since it's not used anymore!

garobrik (approved):
this is super clean, LGTM! just one minor note about the now-unused `sampleAgendaItems.ts` file

Issue #494, opened 2026-08-17
Add existing jupyter notebook


mintybasil:
### Description

We need to commit the [existing notebook]9https://drive.google.com/file/d/1l3Tv52PE_9GmAJ1GYQArVdj3_jeocxHw/view?usp=drive_link) to the repo. 

We should place it in `ml/dev/agenda-item-categorization.ipynb` as it will be a development artifact and not used directly in production.

### To-do (Acceptance Criteria)

- [ ] Notebook committed in `ml/dev/`

Issue #373, opened 2026-03-14
Tagging backend: AgendaItemCategories should get populated right after we fill AgendaItemSubjectTerms


jeromegv:
### Description

This is work that is needed before we bring the tagging system to the frontend

Every hour we bring new agenda item and then populate the table `AgendaItemSubjectTerms` using the subject terms that the city provide in TMMIS

However we need to map the subject terms to the parent categories that will be the effective tagging system. 

The code already exists to do this mapping and have the final categories tied to the agenda item within the table `AgendaItemCategories` (using the categorization we already set in the table `TagCategories`), however this process is not done on a regular basis, it's not part of our update flows in github actions. 

So right now the table `AgendaItemCategories` is missing a lot of agenda items, it's just whatever was there when this code was merged.

So we need to integrate this fully so that AgendaItemCategories stays updated every time we bring new agenda items. And also backfill for the agenda items that were missed.

PR #499, opened 2026-08-17
Implement Modal app for categorization

### LINKED ISSUE #485: Implement Modal app to replicate jupyter notebook
### Description

Currently we have ML inference done via a Jupyter notebook. To move this into production we need to implement a Modal app that replicates the core functionality of the notebook.

The app will be invoked from a GHA workflow roughly like this:
```shell
modal run classify_subject_terms.py::classify_terms \
  --terms-file ml/output/new_terms.json \
  --categories ml/categories.json \
  --write-result ml/output/modal-results/classified_terms.csv
```

This should be an _ephemeral_ app.

### To-do (Acceptance Criteria)

- [ ] Application runs inference task (inference + post-processing)
- [ ] Validation has been performed against jupyter notebook to ensure consistency
- [ ] Terms are provided via a JSON file as input 
- [ ] Categories are provided via a JSON file as input
- [ ] Terms are returned along with their associated categories as csv (format TBD)

mintybasil:
## Description

Resolves #485 

Introduces a [Modal](https://modal.com/) app to replicate the zeroshot classification from the notebook in #494.

We use Modal to offload the classification to their platform, allowing us to automate the ingestion of new subject terms.

Other changes include:
- Duplicates notebook inputs (`categories.json`, `post-processing-rules.json`). This allows us to separate the development from the production version
- Adds randomly selected samples of subject terms to be used for testing (`ml/samples`)
- `compare_outputs.py` - can be used to compare the outputs of the notebook or modal app
- Fixes `docker:start` command

## Testing instructions

Please see `ml/prod/README.md`. You can use one of the sample from `ml/samples`.

## Checklist

- [ ] I have tested these changes in the preview site generated by this PR (you must finish opening the PR first)


docs/SUBJECT_TERM_CLASSIFICATION.md
```diff
@@ -0,0 +1,36 @@
...
+
+4. Run `generateCategoryMappings.ts` to extract data from CSV and merge with other subject term forms
+    - Consumes the classification output CSV
+    - Outputs array of JSON objects, representing each term:
+      - `{"category": string, "tagRaw": string, "tagNormalized": string, "tagSlug": string}`
+
+5. Run `ingestClassification.ts` to insert results into the database
+    - Purges and re-inserts each inputted subject term + category into `TagCategories` table
+    - Updates `AgendaItemSubjectTerms` accordingly
+
```
jeromegv: We also need to repopulate `AgendaItemCategories` because that table will now refer to old categories name, not the new ones.

docs/SUBJECT_TERM_CLASSIFICATION.md
```diff
@@ -0,0 +1,36 @@
...
+## Data Flow
+
+This section aims to describe the flow of data through the various components. All of this occurs automatically inside scheduled Github Actions workflows.
+
+1. Raw data is fetched from TIMMS
+    - The raw agenda items are inserted into the DB table `RawAgendaItemConsiderations`
+    - Table `AgendaItemSubjectTerms` is populated with exploded and normalized forms 
+        - First the raw field is exploded to separate the individual terms
+        - Normalized and slugified forms are constructed and stored for each term
+
```
jeromegv, line 16: Table `AgendaItemCategories` is populated with the agenda item ID and the current category matching the subject terms are matched with
See here if you are curious: https://github.com/civic-dashboard/civic-dashboard-web/issues/373#issue-4076654400

docs/SUBJECT_TERM_CLASSIFICATION.md
jeromegv: (This only applies if we created a new set of tags, which we are for this migration, but not necessarily after if all the tags are the same)

jeromegv (commented):
I approved but still left a few comments

Can you also DM me the modal credentials you have?

docs/SUBJECT_TERM_CLASSIFICATION.md
mintybasil: Good point, added

docs/SUBJECT_TERM_CLASSIFICATION.md
mintybasil: True! Updated the description here, will also need to reflect in the scripts (PR #505). 

Completely rebuilding the table seems expensive, will explore other options.

docs/SUBJECT_TERM_CLASSIFICATION.md
```diff
@@ -0,0 +1,34 @@
+# Subject Term Classification
+
+A **subject term** is a type of keyword that is included in the metadata of agenda items. The total set of subject terms is growing, and their usefulness for querying agenda items is limited. To mitigate this, we have utilized zeroshot classification to associate a category with each subject term.
+
+
+## Data Flow
+
+This section aims to describe the flow of data through the various components. All of this occurs automatically inside scheduled Github Actions workflows.
+
+1. Raw data is fetched from TIMMS
```
garobrik: typo: 
```suggestion
1. Raw data is fetched from TMMIS
```

ml/samples/subject_terms_sample_500.txt
garobrik, line 1: question: i don't think there's a crazy amount of subject terms, any reason not to just check in a full export of the current set that we could update as needed? seems like it would be convenient for dev work

ml/prod/requirements.txt
garobrik, line 1: suggestion: for prod, we should pin dependency versions. also, pyproject.toml is a more modern/flexible dependency specification format which is `uv` native. but not sure if there are environmental constraints on what we can do here.

ml/prod/app.py
```diff
@@ -0,0 +1,172 @@
...
+    for i, rule in enumerate(rules, 1):
+        target_cat = rule["category"]
+        pattern = rule["pattern"]
+        count = 0
+        for row in results:
+            if (re.search(pattern, row['subject_term'], re.IGNORECASE)
+                    and row['max_val'] < 1.0
+                    and row['max_cat'] != target_cat):
+                row['max_cat'] = target_cat
+                row['max_val'] = 1.0
```
garobrik, line 102: question: should we also update `row[target_cat]` to `1.0`?

garobrik (approved):
approved with no blocking comments! this is my first time really engaging with the ml categorization stuff, so also throwing some high level questions/comments in:
- would it make sense to have multiple possible categories per subject term, or is it too easy for that to get noisy? it seems like at least being able to specify that with a post-processing rule would help, and it's not hard for me to imagine a subject term being related to multiple categories 
- I wonder whether it would help if the input to the classification included any sample agenda item content, since the subject terms can be quite brief
- thinking ahead, it would be nice to have an easy way from the UI to see why an agenda item got a certain categorization, so that it doesn't require mucking around in files/the DB
- I'm guessing this all runs on the free $30/mo right now, but what is the API cost of running a round of inference on Modal?

garobrik (approved):
This is my first time engaging with the ml classification stuff so I have some high level questions but this is approved with no blocking comments!
 - would it make sense for subject terms to be able to be mapped to multiple categories even if only from manual post-processing rules? 
- thinking ahead to implementation, it would be nice to have a way from the UI to see why an agenda item received a certain categorization so you don't have to muck around in the database or files to see which subject term led to that categorization 
- I assume this is all in the modal free plan still but what is the API cost of running around of inference or something like this?

mintybasil:
@garobrik thanks for the feedback! Would be great to have more eyes on this, thanks for taking the time to ask questions 🙏 

> - would it make sense to have multiple possible categories per subject term, or is it too easy for that to get noisy? it seems like at least being able to specify that with a post-processing rule would help, and it's not hard for me to imagine a subject term being related to multiple categories
> - I wonder whether it would help if the input to the classification included any sample agenda item content, since the subject terms can be quite brief

Your intuition is right - many terms could be assigned to multiple categories, but the result of that is more noise. 

We are using a zeroshot classifier which only allows 512 input tokens. We could certainly add more context within that limit, but we would need to be selective. Tyna and I are starting to explore using an LLM, first for summaries then expanding into categorization/querying. 

Ultimately, I expect the results from this version will be received as an improvement to the existing tagging system, but still lacking in specificity. More ideas and experimentation are definitely needed.

> - thinking ahead, it would be nice to have an easy way from the UI to see why an agenda item got a certain categorization, so that it doesn't require mucking around in files/the DB

+1 for observability. The question "why did this subject term get assigned this category?" doesn't have a very satisfying answer. Simply put - because the model gave that category the highest score 😅.  

Some additional thoughts:
- It could make sense to store the artifacts from the GHA workflow, in case we want to examine them. 
- We do have the jupyter notebook (`ml/dev/`), which should be able to reproduce the results from prod and allow deeper introspection. It includes some visuals as well to help examine the results. (Note: There can be a small differences in the scores due to a differences in hardware, but that should be negligble)
- Adding a visualization to the "Labs" page could be a cool addition, but I don't expect the raw data to be useful for users

> - I'm guessing this all runs on the free $30/mo right now, but what is the API cost of running a round of inference on Modal?

I was able to run the whole set of 3500+ terms in ~1min for $0.17.

ml/samples/subject_terms_sample_500.txt
mintybasil: These are included so that we can run experiments on smaller sets of data. Running the full set isn't necessarily beneficial as we don't have a way to validate the output, other than manual inspection. 

As a data scientist -> having a small set of terms with a manually created validation set is the most useful. 
As a software dev -> you really want the DB fully populated so you can run end-to-end. Extracting a list of subject terms is cheap.

(I have been tempted to setup a database export to R2, precisely for this purpose. Currently just keeping my own copy on hand.)

ml/prod/requirements.txt
mintybasil: Good suggestions! I'm new to `uv` and have limited python background. My intent has been to push for `uv` but not force it on people. 

I will pin this, and switch to pyproject.toml 👍

garobrik:
ok that all makes sense, thx for the breakdown! only follow-up comment is re: observability: subject term -> category is opaque, but it'd be nice to see which subject terms cause an agenda item to receive the categories it did

-------- Original Message --------

ml/prod/app.py
mintybasil: This is intended to mimic the jupyter notebook. There isn't really any benefit to updating `row[target_cat]`, as the result of the classification is read from `max_cat`/`max_val`. You may also want to preserve the actual score for analysis

jeromegv:
> would it make sense to have multiple possible categories per subject term, or is it too easy for that to get noisy? it seems like at least being able to specify that with a post-processing rule would help, and it's not hard for me to imagine a subject term being related to multiple categories

As context, most agenda items have multiple subject terms, so an agenda item can still end up with multiple categories and we are hoping that those multiple subject terms provide sufficient context.

